PR #1049, opened 2023-01-25
Feature/update corpora

### LINKED ISSUE #954: Update newspaper corpus definitions
As per recent updates, 
- [ ] Corpora can add settings for what constitutes the "context" of a document (e.g. the book it is a page of). Proposed changes in https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/762
- [ ] Corpora can add language analysers for stopword removal and stemming, which unlock new visualisation features. They are not required: visualisations will work fine without them. Similarly, corpora can add a token count field to visualise relative term frequency.
 - [ ] The ngram visualisation is new and has not been added to old corpora. (The term frequency graph has already been added.) Just include it in the list of visualisations for the main content field, if desired. May not be suitable for very large corpora.
 - [ ] Corpora can set a sortable field as the primary sorting field, which will be used to sort documents if there is no query.
 - [ ] If a corpus has a text field where we want both filtering and/or histogram functionality and full text search, this is now possible by mapping it as a keyword field with a text multifield. (See the speaker field in the parliament corpora for an example.)

BeritJanssen:
This branch is built on top of #762, and also adds mappings to update the old corpora to newer standards. Close #954 .


lukavdplas:
@BeritJanssen what is the status of this PR? No hurry, but if you are not expecting to finish it anytime soon I would like to re-open #762, as it can be merged straight away.

BeritJanssen:
True, I kept this PR open until after the Django merge, and then forgot about it. This branch can be merged now - it adds document context to multiple corpora. It cannot hurt to reopen #762 and compare whether this branch misses anything from there -- as far as I remember, I did incorporate changes from that branch but I'm not quite sure.

Issue #1149, opened 2023-06-08
Views for user tag management


lukavdplas:
The tag app in the backend should include views to:

- [x] Add and delete tags for users
- [x] Change a tag's name, description, or colour
- [ ] Return the existing tags by the user on a document - or more practically, a bulk request for a page of documents.
- [x] Return an overview of a user's tags
- [ ] Return an list of all tags a user has used on a particular corpus (can be combined with the one above)

PR #1171, opened 2023-06-22
Feature/tag views

JeltevanBoheemen:
Partially implements https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1149

After discussion we decided to choose M2M relation for documents, because it is expected to speed up querying, and makes it way more flexible. 
See test implementation for suggestions on how to retrieve and store tagged documents.


backend/tag/conftest.py
```diff
@@ -18,15 +25,49 @@ def auth_user_tag(db, auth_user):
...
+    for i in range(DOCS_PER_TAG_LIMIT-1):
+        tagged_doc = TaggedDocument.objects.create(corpus=mock_corpus_obj,
+                                                   doc_id=i)
+        auth_user_tag.tagged_docs.add(tagged_doc)
+        docs.append(tagged_doc)
+    return docs
+
+
+@pytest.fixture()
+def too_much_docs(mock_corpus_obj, near_max_tagged_documents):
```
BeritJanssen, line 69:
```suggestion
def too_many_docs(mock_corpus_obj, near_max_tagged_documents):
```

backend/tag/permissions.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.permissions import BasePermission
+
+
+class IsTagOwner(BasePermission):
+    def has_object_permission(self, request, view, obj):
+        return request.user.is_superuser or (obj.user == request.user)
```
BeritJanssen, line 6: I'm not 100% sure we want to give superusers ownership over all tags. In some projects (People & Parliament, for instance), researchers may be superusers, who may be interested in the tagging feature, but sandboxed in their own tagging system - or does this permission only apply for the admin interface? In there, it may be certainly a nice feature to be able to supervise tags of other users.

BeritJanssen (changes requested):
Nice! Only wondering about giving the superuser permission for other users' tags (see comment).

backend/tag/models.py
```diff
@@ -19,29 +18,20 @@ class Tag(models.Model):
...
-            blank=False,
-            null=False,
-            max_length=512,
-        ),
-        default=list,
-        null=False,
-        size=DOCS_PER_TAG_LIMIT,
+    tags = models.ManyToManyField(
+        to=Tag,
+        related_name='tagged_docs'
```
lukavdplas, line 36: Is is possible to delete the `TaggedDocument` object if the only linked `Tag` is deleted?

backend/tag/signals.py
```diff
@@ -0,0 +1,54 @@
+from django.core.exceptions import ValidationError
+from django.db.models.signals import m2m_changed
+from django.dispatch import receiver
+
+from .models import DOCS_PER_TAG_LIMIT, Tag
+
+
+def tag_exceeds_maximum(tag: Tag, n_added: int):
+    return (tag.tagged_docs.count() + n_added) > DOCS_PER_TAG_LIMIT
```
lukavdplas, line 9: I think I mentioned this before, but this limit is not that useful if we also allow users to
- make a infinite number of `Tag`s
- search for the union of any number of tags.

It could make more sense to impose the limit on a (User, Corpus) scope, rather than (User, Tag), as I think that is more what we care about.

backend/tag/permissions.py
lukavdplas: From what I can tell, this does not affect the queryset in the TagViewset below (which filters on `user=self.request.user`).

That said, this sounds like maybe researchers should not be superusers, and instead be staff users with permissions for specific admin tasks.

lukavdplas: From what I can tell, this does not affect the queryset in the TagViewset below (which filters on `user=self.request.user`).

That said, this sounds like maybe researchers should not be superusers, and instead be staff users with permissions for specific admin tasks.

backend/tag/signals.py
JeltevanBoheemen: making this into an issue

PR #1172, opened 2023-06-22
Automatically verify superuser email adress

JeltevanBoheemen:
Resolves the super-minor inconvenience of having to manually create an `allauth.account.models.EmailAddress` after `django createsuperuser`.


BeritJanssen (approved):
Nice!

PR #1175, opened 2023-06-28
fix range filter crashing page

lukavdplas:
Fixes an issue where loading a corpus with a range filter would completely freeze the page.

### DIFF
--- a/frontend/src/app/filter/range-filter.component.html
+++ b/frontend/src/app/filter/range-filter.component.html
@@ -1,5 +1,5 @@
-<div class="range">
-    <span>{{data.min}} - {{data.max}}</span>
-    <p-slider [animate]="true" [ngModel]="getDisplayData(data)"
-        [range]="true" [min]="min" [max]="max" (onSlideEnd)="update(getFilterData($event))"></p-slider>
+<div class="range" *ngIf="data$ | async as data">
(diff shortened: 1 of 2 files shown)


frontend/src/app/filter/range-filter.component.html
```diff
@@ -1,5 +1,5 @@
-<div class="range">
-    <span>{{data.min}} - {{data.max}}</span>
-    <p-slider [animate]="true" [ngModel]="getDisplayData(data)"
-        [range]="true" [min]="min" [max]="max" (onSlideEnd)="update(getFilterData($event))"></p-slider>
+<div class="range" *ngIf="data$ | async as data">
+    <span>{{data[0]}} - {{data[1]}}</span>
```
JeltevanBoheemen, line 2: Is it a lot of work to convert this to `data.min` and `data.max`?
I understand that for two values the confusion is minor, but for readability it helps to have properties with some semantic information.

JeltevanBoheemen (commented):
Good find, good fix.
Async pipe gang represent. 
I suggest property names for data, see comments.

lukavdplas:
Thanks for the suggestion - I changed it so the template uses the `filter.data` behaviorsubject directly, which has nice property names (and also simplifies the component).

PR #1176, opened 2023-07-04
Feature/about page update

lukavdplas:
Some updates to reflect I-analyzer being public.

- Added a link to the source code in the footer
- Added some references to our github in the about page
- Added a page in the manual about citation that links to our zenodo page. Because I figured most users would be more interested in citing documents than citing I-analyzer itself, this also adds some very basic advice for citing content.
- The text of the about page now has the same size margin on either side.

### DIFF
--- a/frontend/src/app/footer/footer.component.html
+++ b/frontend/src/app/footer/footer.component.html
@@ -22,5 +22,8 @@
             </div>
         </div>
         </div>
+        <div class="block">
+            <a href="https://github.com/UUDigitalHumanitieslab/I-analyzer">Source code</a> (MIT license)
+        </div>
     </div>
 </footer>
--- a/frontend/src/assets/manual/en-GB/citation.md
+++ b/frontend/src/assets/manual/en-GB/citation.md
@@ -8,17 +8,19 @@ If you downloaded a document in I-analyzer, or read it online in the I-analyzer 
 
 For example, a citation might look something like the following:
 
-> Doe, J. (1900, Jan 1). A very interesting article. *The Times.* http://ianalyzer.hum.uu.nl/document/fascinating-corpus/12345678
+> Doe, J. (1900). A very interesting article. *Fascinating Periodical*, *10*(2), 10-13. http://ianalyzer.hum.uu.nl/document/fascinating-corpus/12345678
 
 The specifics will depend on the citation style you are using and the type of data in the corpus.
 
 Documents in I-analyzer have a unique link, which looks like the one in the example above. This is the most appropriate way to reference a specific document.
 
+Note that not all data on I-analyzer is public. For instance, some corpora are only available with an institutional login. This means that your readers may not be able to follow the link to I-analyzer.
+
 ### Citing collection of documents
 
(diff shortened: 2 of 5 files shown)


frontend/src/app/footer/footer.component.html
```diff
@@ -22,5 +22,8 @@
             </div>
         </div>
         </div>
+        <div class="block">
+            <a href="https://github.com/UUDigitalHumanitieslab/I-analyzer">Source code</a> (MIT license)
+        </div>
     </div>
```
BeritJanssen, line 28: Maybe also add a Github icon here?

frontend/src/assets/manual/en-GB/citation.md
```diff
@@ -0,0 +1,36 @@
...
+
+If you want to cite a document that you accessed through I-analyzer, be sure to read the documentation of the corpus. This may include more specific information about the data and how to cite it. This page gives some general advice on citing documents from I-analyzer in your research.
+
+### Citing a document from the I-analyzer interface
+
+If you downloaded a document in I-analyzer, or read it online in the I-analyzer interface, your citation should make it clear that you retrieved it through here. This provides accountability: loading corpora into I-analyzer is a process that may introduce mistakes, so you should always cite the resource that you actually referenced.
+
+For example, a citation might look something like the following:
+
+> Doe, J. (1900, Jan 1). A very interesting article. *The Times.* http://ianalyzer.hum.uu.nl/document/fascinating-corpus/12345678
```
BeritJanssen, line 11: How to deal with the fact that the links will not be accessible to anyone? I don't have an answer, just wonder whether we should include a caveat about this.

frontend/src/assets/manual/en-GB/citation.md
```diff
@@ -0,0 +1,36 @@
...
+
+The specifics will depend on the citation style you are using and the type of data in the corpus.
+
+Documents in I-analyzer have a unique link, which looks like the one in the example above. This is the most appropriate way to reference a specific document.
+
+### Citing collection of documents
+
+For distant reading research, you may want to cite a large collection of documents, rather than a few specific ones. In this case, it is usually most appropriate to cite the corpus as a whole. This may look something like the following:
+
+> Research Software Lab, Centre for Digital Humanities, Utrecht University (2023). *The Times* [data set]. http://ianalyzer.hum.uu.nl/search/times
```
BeritJanssen, line 21: Again, the Times is not a resource all users will have access to.

BeritJanssen (approved):
I love it! Just some minor suggestions, see comments.

frontend/src/assets/manual/en-GB/citation.md
lukavdplas: This is more by way of example, but a public corpus would be clearer. I will update it.

frontend/src/assets/manual/en-GB/citation.md
lukavdplas: Good point. Yes, I will add a comment about it - I don't think we can do much more than that. In the interest of FAIR data, it would be nice if I-analyzer corpora were public, but that is not always possible.

That said, before the current move to open access journals, you could expect that the vast majority of references in an article would not be publicly accessible, so I would say that citing a closed resource is not a problem per se.

PR #1178, opened 2023-07-04
Bugfix/wordcloud stopwords

### LINKED ISSUE #1177: wordcloud does not use stopwords
The wordcloud stopped using stopwords after #1140 , as the wordcloud function was not adjusted to the changes in the corpus class. (It still uses `corpus.language`).

lukavdplas:
close #1177


BeritJanssen (approved):
👍

Issue #1162, opened 2023-06-19
SearchResults model


lukavdplas:
This describes a continuation of #1139. It continues on some of the same ideas, and addresses a certain bloatedness of the `QueryModel`.

Implement a `SearchResults` model. SearchResults more or less represents the state of the search results page. It should contain the following data:

- a `QueryModel`
- settings for sorting, pagination, and highlighting. Could be grouped into a `SearchResultsSettings` model?
- retrieved documents based on the above
- the total number of documents matching the search
- a focused document, which can be `undefined`. This represents the document shown in the popup, if any.

It should handle the following functions:

- Formulate an elasticsearch query by merging  `queryModel.toEsQuery()` with settings for sorting, pagination, and highlight.
- When the QueryModel or sort/pagination/highlighting settings are updated, the model should fetch new results. It should have access to the `SearchService`  to do this.
- Focus and unfocus a document from the results

The SearchResultsComponent basically contains:

- controllers for sorting, pagination, and highlighting
- a view for the documents

Note that sorting and highlighting are currently properties of the `QueryModel`, but they do not belong there, since they are only relevant within the results tab, so they should be moved here.

lukavdplas:
#1307 would more or less close this. What's left is that the results model should be responsible for sorting and highlighting. This is currently handled by the query model, which is not appropriate. I'll make a new issue for that after merging.

Issue #1150, opened 2023-06-08
Tag documents in interface


lukavdplas:
Add tagging interface to the frontend. The search results, document popup and document page should allow users to:

- [ ] Show existing tags on documents.
- [ ] Add or remove a tag for a document

In addition, the search results page should add

- [ ] #1235

PR #1193, opened 2023-07-12
Feature/document model

lukavdplas:
Changes `FoundDocument` from an interface to a class. I added a few methods to the class that were repeated several times across the application.

This update is mostly a preparation for #1162 and #1150. In the case of tagging: as documents can be displayed in multiple components, I wanted the logic to be handled by the model where possible.


BeritJanssen (approved):
Nice! Perhaps add some comments to `FoundDocument.hasContext` to explain the code, this was the function it took me longest to understand.

PR #1201, opened 2023-07-19
add search history settings

### LINKED ISSUE #1145: Make search history opt-in
Somewhat related to #976, which suggests that users should be able to delete their history. Allow users to turn off the search history, i.e. new queries are not stored in the database. Search queries are not incredibly sensitive, but this is quite trivial to add.

lukavdplas:
Adds a settings screen where users can configure whether to enable search history, or delete their existing history. Close #1145


backend/users/models.py
```diff
@@ -18,3 +18,20 @@ def has_access(self, corpus_name):
         # check if any corpus added to the user's group(s) match the corpus name
         return any(corpus for group in self.groups.all()
                    for corpus in group.corpora.filter(name=corpus_name))
+
+
+class UserProfile(models.Model):
```
JeltevanBoheemen, line 23:
```suggestion
class UserProfile(models.Model):
''' User information that is not relevant to authentication.
E.g. settings, preferences, optional personal information.
'''
```

backend/users/serializers.py
```diff
@@ -2,15 +2,32 @@
...
+
+
+    def update(self, instance, validated_data):
+        profile_data = validated_data.pop('profile', None)
+
+        if profile_data:
+            profile_serializer = UserProfileSerializer()
+            profile_serializer.update(instance.profile, profile_data)
 
+        return super().update(instance, validated_data)
```
JeltevanBoheemen, line 30: Elegant approach

backend/users/signals.py
```diff
@@ -28,3 +28,10 @@ def ensure_admin_email(sender, instance, created, **kwargs):
             print(f'Automatically verified email {instance.email} for {instance}')
         except Exception as e:
             print('Failed to automatically verify admin email', e, sep='\n')
+
+@receiver(post_save, sender=CustomUser)
+def create_user_profile(sender, instance, created, **kwargs):
+    if created:
+        UserProfile.objects.get_or_create(
+            user=instance
+        )
```
JeltevanBoheemen, line 37: This combination makes no sense. If a `User` is created, there should not be a `UserProfile` because it's a 1-to-1 relation. Cases where the `get` part is executed instead are clear bugs and should not be passed silently.
```suggestion
    if created:
        UserProfile.objects.create(user=instance)
```

frontend/src/app/models/user.ts
```diff
@@ -15,6 +20,7 @@ export class User {
         public name,
         public isAdmin: boolean,
         public downloadLimit: number = 0, // The download limit for this user, will be 0 if there is no limit.
-        public isSamlLogin: boolean
+        public isSamlLogin: boolean,
+        public enableSearchHistory: boolean,
```
JeltevanBoheemen, line 24: I'm unhappy with the distinction of the `UserResponse` and the `User` class in the frontend. I think I get the reasoning behind it: let the backend make the distinction, it does not matter for the frontend. However, there is a clear separation of responsibility, which I think should be reflected in the frontend. Also because it will only get more complex down the line (if there's more stuff in the profile) to make do the merging/splitting when communicating with the backend.

frontend/src/app/services/api.service.ts
```diff
@@ -299,6 +304,14 @@ export class ApiService {
         );
     }
 
+    /** send PATCH request to update settings for the user */
+    public updateUserSettings(details: Partial<UserResponse>): Observable<UserResponse> {
```
JeltevanBoheemen, line 308:
```suggestion
    public patchUserSettings(details: Partial<UserResponse>): Observable<UserResponse> {
```

frontend/src/app/services/search.service.ts
```diff
@@ -41,21 +41,19 @@ export class SearchService {
...
-            total: results.total,
-            documents: results.documents,
-        } as SearchResults;
+        let resultsPromise: Promise<SearchResults>;
+
+        if (user.enableSearchHistory) {
+            resultsPromise = this.searchAndSave(queryModel, user, request);
+        } else {
+            resultsPromise = request();
+        }
```
JeltevanBoheemen, line 52: I'm not always a fan of ternary operator, but here it saves a lot of lines. Feel free to ignore.
```suggestion
        const resultPromise: Promise<SearchResults> = user.enableSearchHistory ? 
            this.searchAndSave(queryModel, user, request) : request();
```

frontend/src/app/services/search.service.ts
```diff
@@ -84,4 +82,40 @@ export class SearchService {
...
+
+    /** execute a promise while noting the start and end time */
+    private recordTime<T>(makePromise: () => Promise<T>): Promise<[result: T, started: Date, completed: Date]> {
+        const started = new Date(Date.now());
+
+        return makePromise().then(result => {
+            const completed = new Date(Date.now());
+            return [result, started, completed];
+        });
+    }
```
JeltevanBoheemen, line 101: Sweet. This may be useful in other places, could move it to some util file.

frontend/src/app/settings/search-history-settings/search-history-settings.component.html
```diff
@@ -0,0 +1,25 @@
...
+        You can view your saved history on the <a href="/search-history">search history page</a>.
+        It can be used to quickly get back to earlier queries, or to log your research process.
+    </p>
+    <p>
+        Search histories are stored on the I-analyzer server.
+        They are not shared with others, but developers may use search history
+        statistics to assess the level of interest in different corpora.
+        See our <a href="/privacy">privacy statement</a> for more information.
+    </p>
+</details>
```
JeltevanBoheemen, line 17: Really happy with this. Clear, concise, honest.

frontend/src/app/settings/search-history-settings/search-history-settings.component.ts
```diff
@@ -0,0 +1,15 @@
...
+    selector: 'ia-search-history-settings',
+    templateUrl: './search-history-settings.component.html',
+    styleUrls: ['./search-history-settings.component.scss']
+})
+export class SearchHistorySettingsComponent implements OnInit {
+
+    constructor() { }
+
+    ngOnInit(): void {
+    }
```
JeltevanBoheemen, line 13: Tiniest hill to die on.
```suggestion
export class SearchHistorySettingsComponent {

    constructor() { }
```

frontend/src/app/settings/settings.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+    selector: 'ia-settings',
+    templateUrl: './settings.component.html',
+    styleUrls: ['./settings.component.scss']
+})
+export class SettingsComponent implements OnInit {
```
JeltevanBoheemen, line 8: Another hill!

lukavdplas:
> I would rather not see the merging of `User` and `UserProfile` in the frontend, but keep `profile` as a separate class on a `User`. I (think) I can follow your reasons, and also find them valid.

Fair enough, I wasn't sure about this myself. I would like these to be handled in a single view, but `UserProfile` can also just be included as a nested model.

lukavdplas:
> > I would rather not see the merging of `User` and `UserProfile` in the frontend, but keep `profile` as a separate class on a `User`. I (think) I can follow your reasons, and also find them valid.
> 
> Fair enough, I wasn't sure about this myself. I would like these to be handled in a single view, but `UserProfile` can also just be included as a nested model.

Ah, I remembered the implementation incorrectly, so this reply makes no sense. It's already nested.

I can see the merit of having one view for authentication and one for user information, but the current implementation already merges those. I'm not sure that separation should be included in this PR.

frontend/src/app/models/user.ts
lukavdplas: I'm not sure I understand this. The distinction between `UserResponse` and `User` is between the API and the frontend model. I do think it's good practice to separate those representations. But I think this is about the user profile?

PR #1203, opened 2023-07-20
Feature/peace portal adjustments

BeritJanssen:
This branch *should* add iframe support needed for PeacePortal (tbc):
- it leaves out the header and footer
- it posts a `setHeight` message to the parent (in the case of Peace Portal, this is handled by JavaScript on the embedding WordPress site) to assure that the iframe has a (hard-coded) height

In [Peace Portal's search-results-component](https://github.com/UUDigitalHumanitieslab/peace-portal/blob/peaceportal/release/2.5.0/frontend/src/app/search/search-results.component.ts), there is also a `resultsRendered` Output that (according to comments) has been added for iframe support, but as I could find no other component listening to these events, I assume this is an approach that has been abandoned in favour of the `setHeight` message to the parent

### DIFF
--- a/frontend/src/app/app.component.html
+++ b/frontend/src/app/app.component.html
@@ -1,7 +1,7 @@
 <ia-notifications></ia-notifications>
-<ia-menu></ia-menu>
+<ia-menu *ngIf="!iframe"></ia-menu>
 <main>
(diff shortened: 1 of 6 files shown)


JeltevanBoheemen:
Good work!

PR #1206, opened 2023-07-21
Feature/ng modules

### LINKED ISSUE #1134: Remove UserService
The frontend `UserService` still exists, while it has gradually been replaced by `AuthService`. As soon as the remaining authentication features (#66  and UUDigitalHumanitieslab/I-analyzer-django#61) are implemented, remove `UserService`.

### LINKED ISSUE #1197: Use ngModules to group components
I-analyzer has around 70 components and 15 services (some of which are quite bloated and should really be divided up). It would probably help with clarity if we grouped some clusters of related components/services/models into `ngModule`s, so objects that are only relevant *within* the module do not need to be exported to the rest of the app.

lukavdplas:
This PR divides components etc. up into ngModules. This mostly serves to explicate the structure of the application and make a readable `app.module.ts`.

Notes:
- Angular has some options for further modularisation that are not implemented here, such as modules having their own routing and lazy loading, but such steps are enabled by this update.
- The PR also does not do anything else to modularise the application, such as rearranging files or refactoring code. 

close #1197 
close #1134


JeltevanBoheemen (approved):
That cleaned up `AppModule`

PR #1207, opened 2023-07-21
New highlight controller and es highlight functionality

Meesch:
This branch contains a new highlight controller that is clickable instead of an input field and the highlight functionality is taken directly from elasticsearch for corpora that have been indexed using the term_vector: 'positions_offsets' setting on the top layer of the main content field. 

A new option has been added to the main content field mapping in the backend, which lets users select whether they want the main content field to be indexed as such. The new highlight functionality is disabled for corpora that have not been indexed in this way, but because there is no way (that I could find) to query elasticsearch directly to check the term_vector value of the `_source` field, for now the developers will need to manually declare which corpora are compatible with the new highlight functionality in the local_settings. In the future, when all corpora have been indexed in this way, this variable can be removed from the settings. All the files that were changed for this conditionality can be found in commit `5f09943021a292c2342912f86a98739eacaa9b44` Feel free to suggest a more elegant solution!

The elasticsearch highlight functionality allows for the highlighting of stemmed matches and removes the ugly whitespaces in the highlight. 

The button has been styled according to Luka's feedback, but I share the feeling that it is not really in its 'final form'. I think some user input could be useful here, and we might want to do a review of the UI at some point where we address this in a more global way.


frontend/src/app/search/highlight-selector.component.html
```diff
@@ -1,6 +1,26 @@
...
+                            <i class="fa fa-plus" aria-hidden="true"></i>
+                        </button>
+                        <button class="button is-light" (click)="updateHighlightSize('less')">
+                            <i class="fa fa-minus" aria-hidden="true"></i>
+                        </button>
+                    </div>
+                </div>
+            </div>
+            <div class="control">
+                <button class="button highlight-toggle" [class.is-active]="highlight!=0" [disabled]="queryModel.queryText==null||queryModel.queryText==''" (click)="updateHighlightSize(highlight!=0 ? 'off' : 'on')">
```
lukavdplas, line 20: I think the query text could also be `undefined`, not sure. `[disabled]=!queryModel.queryText` would the the quicker way to do this check, I think, but you could also add a `hasQueryText` property to the query model.

If the highlight is disabled, can you add a balloon explaining why?

frontend/src/app/search/highlight-selector.component.ts
```diff
@@ -2,7 +2,7 @@ import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/
 import { QueryModel } from '../models';
 
 
-const HIGHLIGHT = 200;
+const HIGHLIGHT = 100;
```
lukavdplas, line 5: Isn't this an unreachable value? Why is it the default? Do you need a default value at all?

frontend/src/app/utils/es-query.ts
```diff
@@ -86,11 +86,13 @@ export const makeHighlightSpecification = (corpus: Corpus, queryText?: string, h
             pre_tags: ['<span class="highlight">'],
             post_tags: ['</span>'],
             order: 'score',
-            fields: highlightFields.map(field => ({
-                [field.name]: { }
-            }))
+            fields: highlightFields.map( function(field) {
+                return field.displayType == "text_content" && field.positionsOffsets && corpus.new_highlight ? // add matched_fields for stemmed highlighting                    ({ [field.name]: {"type": "fvh", "matched_fields": ["speech", "speech.stemmed"] }}):
+                ({ [field.name]: {"type": "fvh", "matched_fields": ["speech", "speech.stemmed"] }}):
```
lukavdplas, line 91: This is hard-coding the parliament fields.

frontend/src/app/document-view/document-view.component.html
```diff
@@ -41,7 +41,7 @@
             </div>
             <div *ngIf="tabIndex==0">
                 <ng-container *ngFor="let field of contentFields">
-                    <div class="content" *ngIf="document.fieldValues[field.name]" [innerHtml]="document.fieldValues[field.name] | highlight:query"></div>
+                    <div class="content" *ngIf="document.fieldValues[field.name]" [innerHtml]="highlightedInnerHtml(field)"></div>
```
lukavdplas, line 44: Is there a reason why you did not adapt the highlight pipe for this? If so, I think the pipe is unused now and can be deleted.

lukavdplas (changes requested):
Love the new look, this looks really nice!

The frontend logic does not seem to work well:

- When I make any change to the filters, the highlight setting `highlight=200` gets added to the URL. Highlighting should only be specified in the URL if it is a non-default setting.
- Interestingly, I cannot use the date filter in your branch - it always resets to the default value.

backend/ianalyzer/settings.py
```diff
@@ -89,6 +89,10 @@
 # url to the frontend for generating email links
 BASE_URL = 'http://localhost:4200'
 
+# list of corpora that have been re-indexed using the top-level term vector
+# for main content fields, needed for the new highlighter
+NEW_HIGHLIGHT_CORPORA = []
```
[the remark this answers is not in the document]
Meesch: The reason for implementing it this way, on two location is as follows: the corpus definition declares how corpora _should be_ indexed, whereas the settings reflect which corpora _have been_ re-indexed using those settings. It is only for the corpora that have been reindexed that the new highlighter should be turned on. In time, we can get rid of the list in `settings.py`. I realize it is an ugly solution but it is all I could come up with.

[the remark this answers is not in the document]
lukavdplas: That makes sense; I agree that this works as a temporary solution.

I think it is this line which confused me:
https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/3f9b4594bf231a15c0a1509ea50751310011dac9/backend/corpora/parliament/utils/field_defaults.py#L296

Which suggests that you would still be hard-coding this in individual corpus definitions.

backend/addcorpus/es_mappings.py
```diff
@@ -1,16 +1,22 @@
-def main_content_mapping(token_counts = True, stopword_analysis = False, stemming_analysis = False):
+def main_content_mapping(token_counts = True, stopword_analysis = False, stemming_analysis = False, positions_offsets = False):
```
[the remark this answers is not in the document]
Meesch: I called it `updated_highlighting`, since the `True` state would mean that it has updated highlighting capabilities.

frontend/src/app/search/highlight-selector.component.scss
```diff
@@ -6,4 +6,18 @@
...
+
+.highlights-subtitle {
+    text-align: right;
+}
+
+.field.has-addons {
+    justify-content: flex-end;
+
+}
+
```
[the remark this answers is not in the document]
Meesch: `has-text-right` works, but `is-justify-content-flex-end` does not.

frontend/src/app/search/highlight-selector.component.ts
Meesch: removed it!

frontend/src/app/document-view/document-view.component.html
Meesch: The highlight pipe is still used as a fallback option if es provides no highlight options (such as for the legacy corpora), as well as for the manual.

Meesch:
Alright I think I addressed all your comments. Please let me know what you think!

frontend/src/app/search/highlight-selector.component.scss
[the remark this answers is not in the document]
BeritJanssen: The `highlights-subtitle` class can go.

BeritJanssen (approved):
One css class can go, otherwise all seems sensible!

Issue #1219, opened 2023-08-03
Date filter from route bug 


Meesch:
When you try to navigate to a query with a date filter in the route, it will give you a `ExpressionChangedAfterItHasBeenCheckedError` and reset the date filter to the defaults. The error is only shown explicitly in the dev environment because the prod environment suppresses the error message, but the bug exists in both. I will try and tackle this after lunch.

BeritJanssen:
I also found that it's not possible to change the filters from the route anymore.

lukavdplas:
for the date filter, yes, that is the same issue. did you observe this for other filters?

BeritJanssen:
No, this seems to be only a problem for the date filter, indeed.

PR #1224, opened 2023-08-08
Hotfix/date filter route bug

Meesch:
This branch fixes issue #1219 by replacing the defaultvalues in the date filter with the current values of the filter, which means that the filter object does not reset when you refresh the page. Furthermore, a subscription to the filter updates was added to have the dates in the datepicker reflect the most recent changes in the date filter, whether they come from a set filter, a reset filter, or from the route.


frontend/src/app/filter/base-filter.component.ts
```diff
@@ -25,6 +26,12 @@ export abstract class BaseFilterComponent<FilterData> implements OnChanges {
 
     ngOnChanges(changes: SimpleChanges): void {
         if (changes.filter) {
+            if (this.filterSubscription) {
+                this.filterSubscription.unsubscribe();
+            }
+            this.filterSubscription = this.filter.update.subscribe(() =>
+                this.onFilterSet(this.filter)
+            );
```
lukavdplas, line 34: This has the appearance of creating an update loop. It doesn't, because the SearchFilter class will ignore updates with the same data as its current state. It would be nice to add a comment that explains that.

lukavdplas (approved):
Very happy that this is fixed, thanks for looking into it!

I'm not super stoked on using subscribe/unsubscribe this much, but it is consistent with how the component already works, so this is fine.

If you want, you can add a unit test to the DateFilterComponent to see if it reacts to updates in the DateFilter object.

Meesch:
Addressed your comments and added unitttests for the component, althought they are a little basic and don't test the async stuff.

Issue #979, opened 2022-11-25
Compatibility between old and new corpus definitions


lukavdplas:
Following #978 , we now have a new and legacy format to save corpus definitions. Depending on the implementation we prefer, we have two options:

- Converting old corpus definitions to the new database-only format
- Allowing the system to draw corpus definitions from both the database and the python module.

lukavdplas:
Summary of where we're at:

For now, we'll stick with having two methods of entering corpus definitions. If both of those are saved in the database, it doesn't really create issues during runtime.

That means that we'll also accept that, for the time being, python-based definition may support some features that won't be supported in JSON-based definitions.

lukavdplas:
I'm closing this since there is no particular work that needs to be done for this. The work to create a shared database model (which is then serialised for the API) was effectively finished with #1226. Python corpora don't need further adjustments; JSON-based corpora can be added as an additional option. As for that, #978 covers necessary database expansions, and #1410 covers defining and parsing the JSON model.

PR #1226, opened 2023-08-09
Feature/expand corpus model

lukavdplas:
This is a partial implementation of the steps I suggest [here](https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/979#issuecomment-1547540682). The major changes are:

- The database model for corpora is significantly expanded
- Corpus definitions are loaded into the database when starting the server, rather than with every `/api/corpus/` request.
- When possible, interactions with corpora now rely on the database model rather than importing the python class.

The intention here is that the expanded database model can also be filled through another method (e.g. a yaml/json file or a form). As a side benefit, it also makes it easier to do validation.

This change affects the deployment and development workflow. In deployment, the `loadcorpora` command needs to be added to the startup script. In development, the command is included in `yarn start-back`. However, when you change a python corpus definition, you will need to restart the server or run `loadcorpora` manually to see the changes in the interface.

The PR includes a [documentation file](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/f31fead1617cc5023dd5e833c143cdc0064c8add/documentation/Corpus-definitions.md) which may be a good place to start review.


documentation/How-to-add-a-new-corpus-to-Ianalyzer.md
```diff
@@ -103,3 +115,13 @@ Optional flags:
 - `-d` specifies that an existing index of the same name should be deleted first (if not specified, defaults to false, meaning that extra data can be added while existing data is not overwritten)
 
 The start and end date flags are passed on the `sources` function of the corpus (see above). If you did not utilise them there, they will not do anything.
+
+## Validation
+
+The `CorpusDefinition` class has no built-in validation. However, once you start using the corpus, many of the properties defined in the python class will be loaded into the `CorpusConfiguration` database model. This step does include some validation, so it may raise errors. You ran run the import script with `yarn django loadcorpora`. It is also run when you start a development server with `yarn start-back`.
```
BeritJanssen, line 121:
```suggestion
The `CorpusDefinition` class has no built-in validation. However, once you start using the corpus, many of the properties defined in the python class will be loaded into the `CorpusConfiguration` database model. This step does include some validation, so it may raise errors. You can run the import script with `yarn django loadcorpora`. It is also run when you start a development server with `yarn start-back`.
```

backend/addcorpus/save_corpus.py
```diff
@@ -0,0 +1,131 @@
...
+    '''
+
+    exists = Corpus.objects.filter(name=corpus_name).exists()
+
+    if exists:
+        corpus = Corpus.objects.get(name=corpus_name)
+        if corpus.active:
+            corpus.configuration.delete()
+
+def _try_saving_corpus(corpus_name, corpus_definition, verbose = False, stdout=sys.stdout, stderr=sys.stderr):
```
BeritJanssen, line 97: The function name "try" does not sit right with me - it raises questions as to what happens when the attempt fails. Perhaps naming this function `_save_or_skip` or similar may make it more explicit what it does?

BeritJanssen (approved):
Nice move towards administration of corpora in the admin interface. The only small request I have is renaming the functions starting with `_try` to something that is more explicit about conditions and handling of failure of these attempts.

Issue #959, opened 2022-11-18
Improve term frequency estimation


lukavdplas:
When the number of documents exceeds the limit for the term frequency graph, it makes an estimate based on a sample of the data, but this sample is somewhat rudimentary. I've been thinking about how to improve this.

#### The current algorithm

A description of the current algorithm:

The limit of documents to analyse for the entire graph is 5.000. Based on the total number of results, we calculate the ratio that will actually be searched. Say that we search for _democracy_ and find 50.000 hits; then the ratio is 0.2.

The number of documents that are analysed per bin (e.g. per year in the timeline) is proportional to the number of results in that bin. In the example above, if one year had 1000 results, we would analyse 200 documents in that year.

Now we want to estimate the term frequency per bin. For each bin, we know the total number of results $T$, and have a document limit $L$. The analysis uses two types of requests:
- The search results. For speed, we only include the document IDs, not the actual data, and we can limit the number of results.
- When we select a document ID from the results, we can request the term vector for that document to count the occurrences in that document.

(Working with term vectors rather than the flat text is a separate choice, which is not the topic of discussion here.)

The current logic is:
- Request the $L$ most relevant results, and analyse each one.
- For the remaining documents, assume that the search term occurred once in each. This means that we add $T - L$ to the total frequency.

In our example, we will count the frequency of _democracy_ in the top 200 documents, and get 321 occurences. We then assume that it occurred 800 times on top of that, so we estimate a total of 1121 occurences in that year.

There are some reasons for making the algorithm like this:
- We analyse the first results, i.e. results with the highest relevance. These are the most likely to have multiple occurrences of the search term. It is expected that results will often include a long 'tail' of documents with only one match.
- By searching a fixed ratio of the results in each bin, we asssume that the estimation is _proportional_ to the true number of occurences. So when we make an estimate for each year, the results will still show correct proportions between bins.
- It is fairly easy to explain in a manual.

Nonetheless, this method can be improved. Here are some things we might like:

- Improved accuracy
- Get away with analysing fewer documents by making some (informed) assumptions, thus improving speed.

So now the question: we have $T$ results and we are only allowed to analyse $L$ documents. How do we select these documents, and how do we then estimate the total frequency?

#### Thoughts on improvement

Consider what the results look like. Remember that they are ranked by relevance, so we generally[^1] expect the frequency per document to decrease with the ranking.

[^1]: I say 'generally, since: elasticsearch relevance is based on more than just the raw frequency.

[...]

PR #1228, opened 2023-08-09
Feature/term frequency fixes

BeritJanssen:
Instead of fitting a curve over the matches, this branch now goes back to sampling a sizeable batch of the data first, and estimating then by assuming linear decay between the end of the sampling interval and match 1. This means we'll probably overestimate (instead of underestimate) the real number of matches in some cases, but fitting an exponential curve over the data proved not feasible due to fluctuations in the data (see #959). We may still decide that the estimate of decay is not worth it and roll back the changes related so our estimate is back to `sampled_matches + n_not_analyzed_docs`. What do you think, @lukavdplas?

This branch does introduce one essential (frontend) fix though: sorting by relevance rather than date (or other sort, which used to be kept in the query model) when requesting the top hits in the `visualization.term_frequency.get_match_count` function.

### DIFF
--- a/backend/visualization/term_frequency.py
+++ b/backend/visualization/term_frequency.py
@@ -87,8 +87,8 @@ def get_match_count(es_client, es_query, corpus, size, fieldnames):
     mean_last_matches = sum(matches[-ESTIMATE_WINDOW:]) / ESTIMATE_WINDOW
     # we estimate that skipped contain matches linearly decrease 
     # from average in ESTIMATE_WINDOW to 1
-    estimate_skipped = int((mean_last_matches - 1) * skipped_docs / 2)
-    match_count = n_matches + max(estimate_skipped, skipped_docs)
+    estimate_skipped = int((mean_last_matches - 1) * skipped_docs / 2) + skipped_docs
+    match_count = n_matches + estimate_skipped
     return match_count
 
 def count_matches_in_document(id, index, fieldnames, query_text, es_client):
--- a/backend/visualization/term_frequency.py
+++ b/backend/visualization/term_frequency.py
@@ -67,6 +67,7 @@ def get_match_count(es_client, es_query, corpus, size, fieldnames):
     found_hits, total_results = download.scroll(corpus=corpus,
         query_model=es_query,
         download_size=size,
+        client=es_client,
         source=[]
     )
 
--- a/backend/visualization/term_frequency.py
+++ b/backend/visualization/term_frequency.py
@@ -7,7 +7,7 @@ from visualization import query, termvectors
 from es import download
 
 DEFAULT_SIZE = 100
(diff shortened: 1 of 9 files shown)


frontend/src/app/visualization/barchart/timeline.component.html
```diff
@@ -31,8 +31,9 @@
     <div class="message-body">
         <div class="block">
         The results for your search include too many documents to count all matches.
-        Only the first {{percentageDocumentsSearched}}% of documents for each bin are analyzed;
-        the rest are assumed to have only one match for your query.
+        Only a sample of the first {{percentageDocumentsSearched}}% of documents for each bin are analyzed;
+        the rest are assumed to contain linearly decreasing matches between the end of the sample
+        to one match.
```
lukavdplas, line 36: Perhaps just change this into "the match count for other documents is extrapolated" and link to the manual for a full description.

frontend/src/app/visualization/barchart/barchart.directive.ts
```diff
@@ -582,6 +582,13 @@ export abstract class BarchartDirective
         return queryModelCopy;
     }
 
+    /** return a copy of a query model with removed sort parameter (=> relevance sorting) */
+    removeSort(query: QueryModel): QueryModel {
+        const queryModelCopy = query.clone();
+        queryModelCopy.sort.setSortBy(undefined);
```
lukavdplas, line 588:
```suggestion
        queryModelCopy.sort.reset();
```

backend/visualization/term_frequency.py
```diff
@@ -61,22 +64,31 @@ def extract_data_for_term_frequency(corpus, es_query):
...
     skipped_docs = total_results - len(found_hits)
-    match_count = matches + skipped_docs
-
+    if not skipped_docs:
+        return n_matches
+    
+    mean_last_matches = sum(matches[-ESTIMATE_WINDOW:]) / ESTIMATE_WINDOW
+    # we estimate that skipped contain matches linearly decrease 
+    # from average in ESTIMATE_WINDOW to 1
+    estimate_skipped = int((mean_last_matches - 1) * skipped_docs / 2)
```
lukavdplas, line 90: Isn't this:

```suggestion
    estimate_skipped = skipped_docs + int((mean_last_matches - 1) * skipped_docs / 2)
```

lukavdplas (commented):
The code looks good, though I would think the formula for estimating the surface under the curve is different (see my suggestion).

I would think that `auc` would be quite fast since it also uses linear extrapolation, but I trust your judgement on this. The changes proposed here are quite modest so if it improves the accuracy, it's definitely worth it.

backend/visualization/term_frequency.py
BeritJanssen: So it is!

Issue #1235, opened 2023-08-16
 filter widget for tags


lukavdplas:
This should only appear if the user has tagged anything in the corpus. Users can select one or more tags to be used as a filter. The selected tags should be included in the search request.

PR #1236, opened 2023-08-16
Feature/outfactor base filter

lukavdplas:
This is a preparation for #1235 - the idea that there should be a filter based on the user's tags rather than a corpus field.

It separates the `SearchFilter` into two classes: `BaseFilter` and `FieldFilter`. FieldFilter is an implementation of BaseFilter that adds the logic of a filter being tied to a corpus field.

This would allow a `TagFilter` class that is also based on BaseFilter, but which is not associated with a corpus field.


frontend/src/app/models/base-filter.ts
```diff
@@ -0,0 +1,195 @@
+import { ParamMap } from '@angular/router';
+import * as _ from 'lodash';
+import { BehaviorSubject, Observable, Subject } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+/**
+ * Describes shared attributes and properties for filters.
+ */
+export interface FilterAPI<FilterData = any> {
```
BeritJanssen, line 9: Perhaps a better name would be FilterInterface? `API` sounds as though this interacts with the backend.

BeritJanssen (changes requested):
These changes look logical to me, just a small request to find a better name for `FilterAPI`.

PR #1238, opened 2023-08-17
Feature/word model routing

### LINKED ISSUE #849: Search history and routing for wordmodels
The new wordmodels interface does not

- include the query or visualisation parameters in the route
- store queries in the user's search history

These features are less necessary, since word models don't support filters. I can imagine some users would appreciate it, however. It would also make the interface more consistent.

What do we think, is this worth it?

### LINKED ISSUE #1075: Routing for word models
It would be handy if the `param-service` was also used by word models, to keep track of the search term & visualization settings.

BeritJanssen:
close #849 . For the query term comparison, the `term-comparison-editor` now manages its own updates to the route. This branch also fixes problem with several consecutive route updates by buffering and combining updates to a shared subject. @XanderVertegaal , is the use of `bufferTime` in `param.service` a good choice for this?


frontend/src/app/word-models/related-words/related-words.component.ts
```diff
@@ -28,15 +32,37 @@ export class RelatedWordsComponent implements OnChanges {
...
+    }
+
+    teardown() {
+        this.setParams({ neighbours: null });
+    }
+
+    setStateFromParams(params: Params) {
+        if (params.has('neighbours')) {
+            this.neighbours = params.get('neighbours');
+        }
```
lukavdplas, line 61: This would preserve non-default values for `neighbours` if they are removed from the route. If a parameter is included in the route, it should be set from the route in all cases.

```suggestion
        this.neighbours = _.get(params, 'neighbours', 5);
```

frontend/src/app/word-models/related-words/related-words.component.ts
```diff
@@ -28,15 +32,37 @@ export class RelatedWordsComponent implements OnChanges {
...
         }
     }
 
+    initialize() {
+        
+    }
+
+    teardown() {
+        this.setParams({ neighbours: null });
+    }
```
lukavdplas, line 56: This code is repeated across many components.

I recommend adding an abstract `parameterKeys` property the parent class  (in this instance, that would be `['neighbours']`) and handling the reset action in the parent class.

frontend/src/app/word-models/related-words/related-words.component.ts
```diff
@@ -28,15 +32,37 @@ export class RelatedWordsComponent implements OnChanges {
...
 
     ngOnChanges(changes: SimpleChanges) {
         if (changes.corpus || changes.queryText) {
             this.getData();
         }
     }
 
+    initialize() {
+        
+    }
```
lukavdplas, line 52: This has trailing spaces.

```suggestion
    initialize() {
    }
```

frontend/src/app/word-models/word-models.component.ts
```diff
@@ -60,31 +62,48 @@ export class WordModelsComponent implements DoCheck, OnInit {
...
+            this.validateQuery();
+            if (this.queryFeedback === undefined) {
+                this.wordModelsService
+                    .wordInModel(this.queryText, this.corpus.name)
+                    .then(this.handleWordInModel.bind(this))
+                    .catch(() => (this.queryFeedback = { status: 'error' }));
+            }
+            this.activeQuery = this.queryText;
+            this.queryFeedback = { status: 'success' };
+        }
```
lukavdplas, line 90: Similar to above: if the parameters do not have a `query` key, you should not do nothing, but set the state to "no query".

lukavdplas (changes requested):
Nice of have this working!

Unspecified parameters in the route are not handled properly; other than that, it looks fine.

frontend/src/app/param/param-directive.ts
```diff
@@ -2,15 +2,33 @@ import { Directive, OnDestroy, OnInit } from '@angular/core';
...
     constructor(
         public route: ActivatedRoute,
-        private router: Router
-    ) {}
+        private router: Router,
+        private paramService: ParamService
+    ) {
+        this.paramSubscription = this.paramService.bufferedParams.subscribe( params => {
+            if (params.length) {
+                const newParams = Object.assign({}, ...params.map(f => f));
```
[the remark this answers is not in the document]
BeritJanssen: You would think so, but I didn't find one.

frontend/src/app/services/param.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Provides a subject for query parameter changes
+   * Buffers updates in the bufferedParams Observable, subscribed to by ParamDirective
+   */
+
+  public bufferedParams = new Observable<Params[]>();
+  private currentParams =  new BehaviorSubject<Params>({});
+
+  constructor() {
+    this.bufferedParams = this.currentParams.pipe(bufferTime(500));
```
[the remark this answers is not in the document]
BeritJanssen: Unfortunately, the number of components which emit to the subject can change. Indeed, `bufferTime` is not a deterministic way to catch the latest emissions, but in manual tests, works well.

frontend/src/app/param/param-directive.ts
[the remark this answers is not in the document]
lukavdplas: Ok! Can you add a comment explaining what this is supposed to do?

PR #1239, opened 2023-08-17
Feature/functional tests

### LINKED ISSUE #921: e2e test
It seems that access to I-analyzer is flaky with Edge at the moment - it would be good to revisit the e2e tests and add / update tested browsers.

### LINKED ISSUE #1218: Reinstate e2e testing
Removed Protractor due to [deprecation](https://github.com/angular/protractor/issues/5502).
Decide on a new e2e framework and implement it.

BeritJanssen:
This branch adds functional tests (close #921, close #1218), also meant to test the changes from #1238 , so it should be merged after that. For now, the interesting part of the diff is in `functional-tests` (top level). Some issues I ran into:
~~- I followed documentation for implicit / explicit waits as suggested per the [Selenium Documentation](https://www.selenium.dev/documentation/webdriver/waits/) but kept getting `NoSuchElementExceptions`. So for now, there are crude and flaky `time.sleep` statements instead. Does anyone have experience with this?~~ -> solved with seleniumbase
- calls to backend, database, elasticsearch: as obviously, users and corpora aren't guaranteed to be present in other environments ~~, for now, this only "work on my machine".~~
For the second issue, there are two solutions:
1. set up a fake server and set this as a proxy for selenium
2. set up a dedicated Django server, which loads a user, grup & corpus from a fixture, and indexes said corpus based on a small amount of data
Option 2 IMO requires writing less boilerplate code - otherwise, we'd have to create mock responses for calls to `user`, `corpus`, `api/es`, `search_history`, and more. Option 2 would be quite easy to set up in Docker, but then functional tests could only be run with Docker.

A third option -- the path now taken -- is to write tests against a running setup, using an `.ini` file so it can be adjusted for the app instance it's tested against. This is useful, e.g., for running chronjobs on the servers to check everything is working as expected, but also could be added as a sanity check to be run on acceptance to make sure a new feature doesn't have unexpected consequences elsewhere in the application.

For being able to integrate the tests into github actions, we may also still want to add a solution with option 2: a fully dockerized, sandboxed environment that is set up and killed before merging. Leaving this for later, as it requires careful use of caching.


lukavdplas:
>     * calls to backend, database, elasticsearch: as obviously, users and corpora aren't guaranteed to be present in other environments, for now, this only "work on my machine".

This is surprising to me, I would expect a functional test suite to fill in a minimal database, similar to the backend tests, so it can work as a reproducible test of the code. But I'm not familiar with functional testing, is that not typical?

bbonf:
> * I followed documentation for implicit / explicit waits as suggested per the [Selenium Documentation](https://www.selenium.dev/documentation/webdriver/waits/) but kept getting `NoSuchElementExceptions`. So for now, there are crude and flaky `time.sleep` statements instead. Does anyone have experience with this?

I've been using [seleniumbase](https://seleniumbase.io/) for a while. It's a (somewhat monstrous) framework that helps with many of the annoyances in writing selenium tests, and integrates nicely with pytest. Specifically for dealing with waiting, seleniumbase wraps every action and assertion with a (configurable) timeout and that solves most cases.

You can find some examples in excalibur (https://github.com/UUDigitalHumanitieslab/excalibur/blob/develop/functional-tests/test_annotation.py)

BeritJanssen:
Thanks, @bbonf , will check it out!

documentation/Functional-tests.md
```diff
@@ -0,0 +1,25 @@
...
+    password = {your_password}
+
+[corpus]
+    name = {name_of_corpus_your_user_can_access}
+    field = {multiple_choice_filter_field_name}
+    first_option = {first_option_of_that_field}
+    query = {query_term}
+```
+
+Note that in order for all tests to work, the corpus should also have associated *word models*
```
lukavdplas, line 17: This is not a full review, just a comment on a specific point.

This seems highly specific. Especially the last one: a developer who does not work on word models should not have to download word models to run functional tests.

My intuition would be that the current setup of tests is best understood as a test of whether the active environment is "valid" - if there are no word models, nothing word-modelly needs to be tested.

The more generalised version of that is to not include any of this information; instead, let the functional tests explore the interface to assemble a list of available corpora, then open the link of each corpus to see if it works, check if there is a link to word models, see if that works, etc. That is more work than including a lot of assumptions though.

BeritJanssen: Agreed, the word model tests could also just be skipped if they're not present.

As for the assumptions - I think this is something we cannot completely abstract away. For instance, if we trigger the first filter of the first corpus, what values do we expect in the route? I'm also still struggling with this, but in essence, from what I understand, functional tests are built against a *running* application, and not against a sandboxed environment. We discussed the latter for GitHub actions, but I think both routes have merit. I believe if we filled in some of these "assumptions" for one of the live servers, we could get much better diagnostics of whether the server is behaving as it should.

documentation/Functional-tests.md
```diff
@@ -0,0 +1,25 @@
+# Functional test suite
+Functional tests are located in the folder `functional-tests`.
+
+To run functional tests, make sure you have Selenium and Seleniumbase installed (from backend requirements). Also, create an .ini file in the `functional-tests` folder, which should have the following setup:
```
lukavdplas, line 4: Small point, but I imagine `.ini` is not the full filename

lukavdplas (changes requested):
Re. your question if this is ready to merge:

I would say it looks fine, but it's still a problem to me that the tests will not work without a corpus with word embeddings. I usually don't have word models in my local environment - nor do we have those on every server. So that is something that I would want to see changed before merging.

I also have my reservations about a corpus with a multiple choice filter being required, but that is more a matter of principle. Multiple choice filters are technically optional, but we use them in almost every corpus so this is unlikely to be a real issue.

documentation/Functional-tests.md
lukavdplas: Hm, fair enough, these assumptions do make things easier. Still, I think you should be able to configure that features like word models are not present in your environment.

>  Then again, I don't agree that a "valid" setup is what we want to test here - it's that every part of the application reacts as it should. So that if we change one component, we can catch consequences to unrelated components, such as the word models, which may otherwise remain undetected.

> functional tests are built against a running application, and not against a sandboxed environment

This sounds contradictory to me, at least for an ostensibly modular / configurable application like I-analyzer. You can test whether a running environment is working as it should, or you can setup a sandbox that includes every module you want to test.

In the case of the first, you should be able to validate an environment that opts out of some I-analyzer features (such as not having word models or not having multiple choice filters), since that is still a working environment - and yes, that environment will still pass all tests if the word models are broken, because what you are testing is if *this* environment works as it should.

In the case of the second, you can actually make sure that any module is present and tested because you control the whole environment - but with obvious limitations because you are not testing a "real" environment.

I'm somewhat worried that making strong demands like this in the functional tests will make it harder to maintain I-analyzer as a modular application. This would especially get problematic if we would use a similar setup for things like each filter type, term frequency normalisation, document context, tagging, NER, etc.

PR #1242, opened 2023-08-17
Feature/filter tags in more views

### LINKED ISSUE #1234: visualisations / downloads with tags
continuation of #1191 

It is now possible to include tags in a `/_search` request, but the option for tags should also be implemented for other request that use the same query model - namely visualisations and search results downloads.

lukavdplas:
All visualisation and download views now support a `tags` parameter. (close #1234 )

Requests can be formatted like this:

```json
{
    "es_query": {...},
    "tags": [47],
    ...
}
```

where the backend will then insert the proper filter in the elasticsearch query.

The search request is slightly adjusted so it uses the same format.

I've made some adjustments to the typing of such requests in the frontend, so `tags` is recognised as optional parameter in request objects, thought it is not actually used.


backend/es/views.py
```diff
@@ -20,33 +20,19 @@ def get_query_parameters(request):
...
+    are combined with OR logic: a document must match any of them.
 
-    The request data should be the query JSON. You can optionally include
-    extra key/value pairs as query parameters in the request. They
-    will be merged with the JSON data.
-
-    On top of any properties included in the query DSL, the request may also specify
-    a `tags`property. This should be an array of values, referring to IDs of tags
-    on which the results should be filtered.
+    You can optionally include extra key/value pairs as query parameters
```
JeltevanBoheemen, line 33: I think we should point out that these MUST be ES valid keys/values. Is this validated?

backend/tag/filter.py
```diff
@@ -1,5 +1,21 @@
 from tag.models import Tag, TaggedDocument
 from visualization.query import add_filter
+from addcorpus.permissions import corpus_name_from_request
+
+def handle_tags_in_request(request, parameter_key = None):
```
JeltevanBoheemen, line 5: Neat function. Clear way to handle it (y)

JeltevanBoheemen (approved):
This looks excellent, as always.

backend/es/views.py
lukavdplas: Good point, I'll add it to the docstring. As for validation: elasticsearch obviously validates it so the request will fail if you add strange values. I don't think we need to add our own validation layer unless we want to block some queries that are allowed by elasticsearch.

Issue #687, opened 2022-04-19
Enable full data download for wordcloud


lukavdplas:
The `load more data` in the wordcloud does _something_ (shows a loading spinner for a while, then finishes), but perhaps the loaded data is not actually used. In any case, the frequency table is not updated when loading more data. The wordcloud picture is re-drawn, but I don't think the data changes. (It has a random component, so it's hard to see.)

PR #1243, opened 2023-08-24
remove unused code

lukavdplas:
Removes various bits of unused code, namely:

- Corpus definitions for TML and spectators - I don't think these ever made it into production
- The async view for the wordcloud, which does nothing (c.f. #687) 
- The dutch version of the manual, which cannot be accessed and is not actively maintained


frontend/src/app/services/api.service.ts
```diff
@@ -140,11 +140,6 @@ export class ApiService {
         return this.http.post<AggregateResult[]>(url, data).toPromise();
     }
 
-    public wordcloudTasks(data: WordcloudParameters): Promise<TaskResult> {
```
BeritJanssen, line 143: We may still bring back the feature to generate full word clouds, though. So perhaps don't get rid of the related tasks. Or is this now done via another mechanism?

frontend/src/assets/manual/nl-NL/manifest.json
BeritJanssen, line 1: I'm not sure we should just chuck out the dutch translations like that. We may still use them later on. Or did you want to use another localization mechanism?

BeritJanssen (changes requested):
I applaud getting rid of unused corpus definitions, but I'm on the fence about the word cloud task (weren't we going to bring that back?) and the Dutch interface translations (there were plans to actually use these in the future, but perhaps there are other solutions to this that I'm not aware of). While it's easy to bring them back, it's also easy to forget they were ever there once they're gone.

frontend/src/app/services/api.service.ts
lukavdplas: I think if we want to introduce this, it would be by extending this view instead:
https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/842919a4b2bdaa0aee15c85cfb0d81d90ef4c05c/backend/download/views.py#L91-L114

But also, we use git for a reason: we can always revert the commit if we want to reuse this approach.

frontend/src/assets/manual/nl-NL/manifest.json
lukavdplas: No, this was just because it is not actually accessible or maintained.

In the case of documentation: I find this a bit tricky because keeping stale documentation around for years means it becomes increasingly unlikely that it will still be useful. That said, unlike code, this is not really "in the way" as much, so if you want to keep it, I'm fine with that.

BeritJanssen: Let's keep it, because it does have some relevant documentation for the query syntax still, I believe.

frontend/src/app/services/api.service.ts
BeritJanssen: In that case, this can indeed go.

PR #1245, opened 2023-08-31
fix deployment for NEW_HIGHLIGHT_CORPORA

BeritJanssen:
This branch is improperly named hotfix - since the highlight PR was not yet released, merging into develop will do for now, and will ensure that we can deploy even if we don't define the `NEW_HIGHLIGHT_CORPORA` variable in the settings.

### DIFF
--- a/backend/addcorpus/corpus.py
+++ b/backend/addcorpus/corpus.py
@@ -11,12 +11,16 @@ import bs4
 import csv
 import sys
 from datetime import datetime
-from langcodes import Language, standardize_tag
 from os.path import isdir
+
+from django.conf import settings
+from langcodes import Language, standardize_tag
+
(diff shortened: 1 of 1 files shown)


BeritJanssen:
@Meesch : probably a good idea to already add this variable to the deployment module though, with a list of which corpora have already been updated.

Meesch:
> @Meesch : probably a good idea to already add this variable to the deployment module's settings though, with a list of which corpora have already been updated.

So to be clear: I should add the NEW_HIGHLIGHT_CORPORA to the settings in the ianalyzer branch of the deployment repo and update that as corpora are re-indexed?

BeritJanssen:
Yes, or more precisely: the ianalyzer-django branch.

Issue #1247, opened 2023-08-31
Memory leaks in subscriptions to queryModel


BeritJanssen:
Every component that subscribes to the `queryModel.update` subject does so in a `onChanges` trigger. Is this not overkill? Would it not be sufficient to subscribe *once* in the constructor? I believe then we wouldn't even need change detection on the input anymore -- in fact, why listen for updates on the `@Input` and the `queryModel.update` Subject both?

What is more, I don't see any `unsubscribe()` methods in `onDestroy`. Does this not entail that we will subscribe multiple times and never unsubscribe?

@lukavdplas , perhaps there are parts of the implementation that take care of this, in which case it would be nice to document this somewhere.

lukavdplas:
Regarding `onChanges`: I did initially use `onInit` (the constructor is called before receiving inputs), but this would only work if you were sure that the queryModel never going to be re-instantiated. This does happen, when navigation is triggered by the browser (e.g. with back/forward). So `OnChanges` is appropriate. I'm all for documenting this, though I'm not sure where.

You are right about the leakiness, that should be addressed: a lot of these subscriptions are not properly closed when the component is destroyed.

There is also a potential leakiness when the query model is re-instantiated (if the component is still subscribed to the old query model). The "old" query model is always discarded, so this is unlike to cause issues, but it is sloppy. I *think*  the neat way to handle this would be to complete observables on the old query model before making a new one.

As an aside, I would like to get away from all these `subscribe` calls: most components could just work with a piped observable instead.

BeritJanssen:
For now, we could just create a `subscription` member on a component which describes to the query model, and in the `onChanges` call do 
```
this.subscription.unsubscribe()
this.subscription = this.queryModel.onNext.subscribe()
```
and be sure to add a `onDestroy` method with `this.subscription.unsubscribe()`.

BeritJanssen:
Still, I'm not quite clear why we the `@Input` when we get updates on the query model via the observable?

lukavdplas:
> For now, we could just create a `subscription` member on a component which describes to the query model, and in the `onChanges` call do
> 
> ```
> this.subscription.unsubscribe()
> this.subscription = this.queryModel.onNext.subscribe()
> ```

I would prefer a setup like this one that uses `takeUntil` instead of  calling `unsubscribe`.

[...]

PR #1248, opened 2023-08-31
close #1246 set highlight as switched off

### LINKED ISSUE #1246: BUG: route resets after switching visualization view
To reproduce:
- select corpus
- search for a term
- switch to visualization tab
- try to select different visualization

The visualization will keep resetting itself to `resultscount`.
Hold releases until fixed.

BeritJanssen:
This seems to fix bug #1246. Reviewing this code, I realized that the setup with two ways of communicating changes of the queryModel to components, one through a `queryModel.update` Subject, and one through an `@Input` for `search-component`'s children, is rather hard to grasp. Worth revisiting (see #1247 ).


Meesch:
This works, but when the user switches back to the search results, highlighting remains turned off. I think we should merge this, as it fixes a bug that is far more annoying, but we should also think critically about how we want to implement the highlight settings. Let's discuss this in our next meeting?

Meesch (approved):

PR #1252, opened 2023-09-07
Bugfix/es settings

### LINKED ISSUE #1251: Index settings incorrect
Running an index with the `-p` flag will incorrectly create an index with 1 shard and 1 replica (should be: 5 / 1). Should be fixed, *especially* before running a big indexing job from source. @Meesch

BeritJanssen:
close #1251


functional-tests/latest_logs/test_routing.test_filter/basic_test_info.txt
lukavdplas, line 1: This (and everything else from the directory) looks like it have been in gitignore.

backend/es/es_index.py
```diff
@@ -44,8 +44,7 @@ def create(client, corpus_definition, add, clear, prod):
             sys.exit(1)
 
         logger.info('Adding prod settings to index')
-        index_settings = settings.get('index', {})
-        index_settings.update({
+        settings['index'].update({
             'number_of_replicas' : 0,
```
lukavdplas, line 48: Shouldn't we use `number_of_replicas: 1` in production? 🤔

lukavdplas (changes requested):
Ignoring the functional test output, this looks good!

BeritJanssen:
Ah, I thought I'd been careful not to commit those files, but seems I hit the "commit all" button at some point.

backend/es/es_index.py
BeritJanssen: Only after the indexing procedure is complete. That happens at the end of the indexing script.

lukavdplas (approved):

PR #1261, opened 2023-09-13
Feature/filter module organisation

lukavdplas:
Quick update to do 2 things:

- Create a `FilterBoxComponent` which sits at a level between the filter manager component and components for filter controls (e.g. date filter component). Does nothing, but makes the filter manager component a bit more readable.
- Move some filter components into subdirectories to `/frontend/src/app/filter` is less cluttered.

Contains no original code, so no elaborate review is needed. Feel free to approve without going through the code if these suggestions sound good to you.


BeritJanssen (changes requested):
I like it! I'm just wondering whether we might distinguish between filter types (i.e., boolean, date, multiple choice and range) and filter utilities? Then filter-box and ad-hoc-filter might better live on the higher level, so for somebody new to the project it doesn't seem as if they are two more filter types. Just a suggestion though - any organization in this directory is better than no organization, and factoring out the wrappers of the filter widget itself is also definitely a good move.

lukavdplas:
We could move the filter controls into another subdirectory. The ad-hoc-filter should go in there too though, it exists on the same level as boolean filter / date filter /etc

Issue #1085, opened 2023-03-18
Routing: Reorganize visualization options


BeritJanssen:
Currently, ngrams options add 6 query parameters in the route, some of which (e.g., size, position) might also be used as field names on corpora, in which case they'd clash with associated filters. Also, the route becomes very long that way. It might be better to reorganize the visualization options such that the query parameter for ngram becomes `?ngramOptions=s:2,p:any...` This means we'll need to include some functions to get / set the visualization options correctly from and to the route.

PR #1266, opened 2023-09-13
Feature/corpus model validation

lukavdplas:
Add more validation functions to corpus models. This includes a few "soft checks" where the validator raises a warning instead of an error.

I also corrected one genuine error in the definition of DBNL.

addresses [this point](https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1085#issuecomment-1475822878)


backend/addcorpus/filters.py
```diff
@@ -31,6 +34,8 @@ class DateFilter(Filter):
     Filter for datetime values: produces two datepickers for min and max date.
     '''
 
+    mapping_types = ['date']
```
JeltevanBoheemen, line 37:
```suggestion
    mapping_types = (MappingTypes.DATE,)
```
Same goes for rest of the mapping types. Tuple suggestion because this seems like immutable data (though single value tuple always looks a bit weird).

backend/addcorpus/validators.py
```diff
@@ -8,3 +25,147 @@ def validate_language_code(value):
...
+
+def validate_name_is_not_a_route_parameter(value):
+    '''
+    reject names that are also used as query parameters in frontend routes.
+
+    This would create serious bugs in the frontend as those parameters will also
+    be interpreted as filter settings for the field.
+    '''
+
+    forbidden = [
```
JeltevanBoheemen, line 101: Maybe move these to constants?

JeltevanBoheemen (approved):
Nice work, well documented and tested.

lukavdplas:
Labelled "affects elasticsearch index" because the DBNL mapping is adjusted and requires re-indexing. This should be possible without extracting from source data.

PR #1267, opened 2023-09-21
Feature/parallelize ngrams

### LINKED ISSUE #1005: Parallellise ngram graph
Following #962, we could also apply parallellisation to the ngram graph.

The process would roughly need to be divided in to three tasks:

- Assemble the time bins
- Create a frequency list of all relevant ngrams within a time interval <- run this in parallel for all time intervals
- Calculate top n ngrams over the whole period and format the result

BeritJanssen:
Close #1005 .


lukavdplas (approved):
Nice! :smile: 

Note: I have not tried running this code, but I trust that you did. The code looks good (pending the failing test)!

A small nitpick on code quality: there are a lot of one-letter abbreviations as variable names (e.g. `for r in results` vs. `for result in results`), which I personally find detrimental to readability.

PR #1270, opened 2023-09-21
Feature/corpus descriptions

lukavdplas:
- Adds/updates descriptions for the ECCO, Times, Guardian/Observer, and Periodicals corpora - thanks to our colleagues at the university library.
- Merges the newspapers/periodicals categories for corpora. This is a suggestion from Arja: older media don't have a clear distinction between newspapers and magazines.
- Changes the image for the Times corpus - I thought the current image (of the palace of westminster) would be confusing if times and parliament-uk are active in the same environment.


BeritJanssen (approved):
Nice!

PR #1271, opened 2023-09-27
Feature/finland old

BeritJanssen:
Corpus definition for the debates of the pre 1905 Finnish parliament.


backend/corpora/parliament/utils/field_defaults.py
```diff
@@ -250,21 +250,6 @@ def subtopic():
...
-        display_name='Subject',
-        description='Subject that the speech is concerned with',
-        es_mapping=keyword_mapping(),
-        search_filter = MultipleChoiceFilter(
-            description='Search only in speeches about the selected subjects',
-            option_count=50
-        ),
-        visualizations = ['resultscount', 'termfrequency']
-    )
-
```
BeritJanssen, line 267: Duplicate.

backend/corpora/parliament/description/finland-old.md
```diff
@@ -0,0 +1,4 @@
+Minutes from the debates of the multicameral Diet of Finland. The chambers - Nobility, Clergy, Bourgeoisie, and Peasants - met at irregular intervals between 1863 and the Parliament Act of 1906, which founded the modern unicameral Parliament of Finland.
+
+#### Language
+Swedish and Finnish
```
lukavdplas, line 4: This is a bit superfluous since it will be listed at the top of the page.

lukavdplas (approved):
👍

backend/ianalyzer/settings.py
lukavdplas, line 1: What is the use of adding these log settings on develop? (genuine question)

BeritJanssen: There was no logging whatsoever at this moment when indexing.

lukavdplas: I did notice these settings were missing in the deployment module (https://github.com/UUDigitalHumanitieslab/deployment/issues/38), but this is `settings.py`, so it will not affect production. For development environments, I don't really see the use of an indexing log to be honest.

BeritJanssen: It's not necessary that this is a file logger indeed. I changed that. But it is definitely necessary to capture the logging from indexing, also during developments.

PR #1272, opened 2023-09-27
fix getting word model query from route

### LINKED ISSUE #1269: word models broken
When entering a query in the word models pane, the route will change, but then nothing happens.

lukavdplas:
close #1269


lukavdplas:
Ah, forgot to add it. It does!

BeritJanssen (approved):
Phew, nice catch! I thought this was going to be some nightmarish routing spaghetti.

PR #1273, opened 2023-09-28
Feature/n gram params

### LINKED ISSUE #1085: Routing: Reorganize visualization options

BeritJanssen:
Close #1085 . The branch also includes a modification of the Dockerfile for the frontend, and provides the `--no-sandbox` option when running the frontend tests via Karma - this is necessary for the containerized environment. @oktaal is this setting a problem in other environments?


frontend/src/app/models/visualization.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+            'any',
+            false,
+            'none',
+            50,
+            10,
+            'date'
+        );
+    });
+
+    it('should convert themselves to a param string', () => {
```
lukavdplas, line 18:
```suggestion
    it('should convert itself to a param string', () => {
```

frontend/src/app/models/visualization.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        const paramString = ngramParameters.toRouteParam();
+        expect(paramString).toEqual(
+            's:2,p:any,c:false,a:none,m:50,n:10,f:date'
+        );
+    });
+
+    it('should set themselvers from a param string', () => {
```
lukavdplas, line 25:
```suggestion
    it('should set itself from a param string', () => {
```

frontend/src/app/visualization/ngram/ngram.component.ts
lukavdplas, line 1: The method `parametersKey` in this component also converts an `NgramSettings` objects to a string - you should be able to replace it with your new function.

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -64,6 +63,15 @@ export class NgramComponent extends ParamDirective implements OnChanges {
...
         super(route, router, paramService);
+        this.currentParameters = new NgramParameters(
+            this.sizeOptions[0].value,
+            this.positionsOptions[0].value,
+            this.freqCompensationOptions[0].value,
+            'none',
+            this.maxDocumentsOptions[0].value,
+            this.numberOfNgramsOptions[0].value,
+            'date'
+        );
```
lukavdplas, line 74: I would expect the initial values to be based on the route. Does this work when loading the page with parameters in the route?

BeritJanssen: Hm, good point!

lukavdplas:
Almost forgot: https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/b1dc35f7ede3ead6e3836d282506cccd5e842dab/backend/addcorpus/constants.py#L36 is outdated in this PR, please update that as well.

lukavdplas:
A thought: remapping the parameters like this is technically a breaking change, as bookmarked links to visualisations will cease to work. It may be possible to add a guard or resolver that parses or redirects legacy links.

Anyway, I leave that to your judgement. The ngram graph is only active for the parliament corpora and troonredes, so I would only do this if the update would be a disruption to the P&P researchers.

frontend/src/app/visualization/ngram/ngram.component.ts
BeritJanssen: I checked once more to be sure, and indeed, it works as I'd intended: the constructor sets the defaults, and later in the component's lifetime, the route settings will be applied, if present.

BeritJanssen:
I will check indeed whether anyone saved links to this visualization - but I suspect that the option to bookmark visualizations isn't really used.

backend/addcorpus/constants.py
```diff
@@ -41,13 +41,6 @@ class VisualizationType(Enum):
...
     'visualizedField',
     'normalize',
-    'size',
-    'positions',
-    'freqCompensation',
-    'analysis',
-    'maxDocuments',
-    'numberOfNgrams',
-    'dateField',
 ]
```
lukavdplas, line 44: `ngramSettings` should now be *added* as a claimed parameter.

Issue #1282, opened 2023-10-06
Direct downloads do not get filename from backend


lukavdplas:
For the "direct download", i.e. search results under 10.00 documents, this line is supposed to get the filename:

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/129f68f09f9081e7eb5831f94b9f595bc9f4ed6e/frontend/src/app/services/download.service.ts#L40

This does not work because

- The backend does not send a `filename` header. The filename is included in `content-disposition`
- Not all response headers are exposed to javascript, so even if you add a `filename` header in the response, it will not be read.

It looks like FileSaver handles this by turning the filename into `download.csv`.

Issue #1181, opened 2023-07-06
Download fails with long url


lukavdplas:
I tried to download the search results from [this query](https://ianalyzer.hum.uu.nl/search/goodreads?language=Afrikaans,Albanian,Arabic,Bengali,Bulgarian,Catalan,Chinese,Croatian,Czech,Danish,Dutch,Estonian,Finnish,French,German,Hebrew,Hindi,Hungarian,Indonesian,Italian,Japanese,Korean,Latvian,Lithuanian,Macedonian,Malayalam,Modern%2520Greek%2520(1453-),Nepali%2520(macrolanguage),Norwegian,Persian,Polish,Portuguese,Romanian,Russian,Slovak,Slovenian,Somali,Spanish,Swahili%2520(macrolanguage),Swedish,Tagalog,Tamil,Thai,Turkish,UNKNOWN,Ukrainian,Urdu,Vietnamese,Welsh) but the download failed. 

Note the long URL; shorter queries worked fine. The celery log includes 

```
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(254)
```

Which is prorobably the reason

lukavdplas:
Resolved by #1283

PR #1283, opened 2023-10-06
Feature/csv filenames

### LINKED ISSUE #1052: csv filenames on server
The csv filenames on the server are a bit awkward. One issue is that the long filenames are not very clear to users, see #937

Another issue that these filenames are supposed to be unique. Since all csv files are stored in the same directory, non-unique filenames would mean that csv start overwriting one another - hence why the filename needs to completely describe the content. For search results, this is not quite true (I mention in #891 that the download limit is not represented, an problem that is not actually resolved), but close enough. For visualisations, an actual description would be a lot longer, since it needs to capture the query as well as the parameters. The uniqueness problem is currently solved somewhat awkwardly by adding a timestamp to the filename.

Anyway, now that we have a downloads table, it would better to just store the download with just the ID as the name, e.g. `47.csv`. Then the more descriptive filename for the user can be generated from the datarow in the table.

lukavdplas:
The backend now saves CSVs with their ID as the filename, e.g. `123.csv`. (close #1052 )

Also adjusts the filename sent to the user to a single pattern: all downloads are named `{download type} {corpus} {timestamp}.csv`, e.g. `search_results dbnl 2023-10-06T13:42.csv`. (However, due to #1282, this only affects task-based downloads.)

I think this would also address #1181 but I'll need to deploy the branch to check.


backend/download/tests/test_full_data.py
```diff
@@ -24,7 +24,7 @@ def test_timeline_full_data(small_mock_corpus, index_small_mock_corpus, small_mo
 
     group = tasks.term_frequency_full_data_tasks(full_data_parameters, 'date_term_frequency')
     results = group.apply().get()
-    filename = tasks.make_term_frequency_csv(results, full_data_parameters)
+    filename = tasks.make_term_frequency_csv(results, full_data_parameters, 0)
```
BeritJanssen, line 27: It would be easier to parse the parameters in the tests if `log_id` was explicitly named here.

backend/download/views.py
```diff
@@ -51,13 +52,13 @@ def post(self, request, *args, **kwargs):
...
-            directory, filename = os.path.split(csv_path)
-            # Create download for download history
             download = Download.objects.create(
                 download_type='search_results', corpus=corpus, parameters=request.data, user=request.user)
+            csv_path = tasks.make_csv(search_results, request.data, download.id)
+            directory, filename = os.path.split(csv_path)
+            # Create download for download history
             download.complete(filename=filename)
-            return send_csv_file(directory, filename, 'search_results', request.data['encoding'])
+            return send_csv_file(download, directory, filename, 'search_results', request.data['encoding'])
```
BeritJanssen, line 61: Do we need to pass `filename` and `download_type` if they are part of the `download` object?

BeritJanssen (changes requested):
Nice! Just a couple of small comments, the main one being whether `backend.download.send_csv_file` could do with fewer arguments,  if the `download` object already contains the `filename` and `download_type` of the csv file.

BeritJanssen (approved):

PR #1293, opened 2023-10-17
Add issue templates

lukavdplas:
Small suggestion to add issue templates to the repository.

I've added a "feature request" one (which is pretty much the github default) and a "bug report" (which has some more I-analyzer specific content).

This is mostly to provide some guidance and clarity for users / researchers interested in contributing, but they are also intended for developers. (I notice we often forget to include necessary context in bug reports.)

Not something we've discussed, feel free to reject the idea of adding templates at all!


BeritJanssen (approved):
Good idea! I do wonder if it's going to get in the way when we open feature / bug issues ourselves though, but I'm all for trying. We can always change it again.

PR #1296, opened 2023-10-23
Use chart.js for wordcloud

lukavdplas:
Wordcloud is now based on chart.js, rather than d3. Also updates chart.js to version 4.

This is mostly to simplify the code base. The chart.js implementation is more high-level and similar to other charts. It does add one feature: the wordcloud now has tooltips.

Screenshot:

screenshot of wordcloud graph


frontend/src/app/visualization/wordcloud/wordcloud.component.ts
```diff
@@ -77,6 +77,10 @@ export class WordcloudComponent implements OnChanges, OnInit, OnDestroy {
         }
     }
 
+    ngOnDestroy(): void {
+        this.chart.destroy();
```
BeritJanssen, line 81: The word cloud tasks aren't aborted now anymore. Or is this done without Celery now?

BeritJanssen (commented):
Nice, but I do wonder why the mechanism to cancel tasks is gone. In practice, the requests are pretty fast anyway, but it would still be neater to clean up the tasks when the user switches to another view before the task has finished on the backend.

lukavdplas:
The wordcloud doesn't create celery tasks anymore (nor does the component keep track of any) - I forgot to remove these lines in #1243

BeritJanssen (approved):

PR #1298, opened 2023-10-25
Hotfix/saml uu authorisation

### LINKED ISSUE #1277: Students no longer have access to paid corpora
The two students in the I-A workshop did not have access to the non-public corpora, even though they should have access. Potentially something that went wrong with the migration?

lukavdplas:
Fixes an issue where SAML users did not receive proper authorisation.

close #1277


BeritJanssen (approved):
Ah, good catch!

PR #1306, opened 2023-11-02
Feature/save search history from backend

lukavdplas:
Moves responsibility for logging search history to the backend.

The backend now saves queries when executing the `_search` view. The frontend no longer makes requests to save search history.

I made this change to help with simplifying state management of the frontend in #1307 . It also just seems more appropriate to let this be handled by the backend.

This changes the logic to avoid logging many similar queries at the same time. This was based on the context of the request in the frontend. The new version checks for very similar queries in the user's recent history.


backend/api/save_query.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime, timedelta
+
+def save_query(user, es_query):
```
JeltevanBoheemen, line 3:
```suggestion
def should_save_query(user, es_query):
```

backend/api/save_query.py
```diff
@@ -0,0 +1,43 @@
...
+
+def has_aggregations(es_query):
+    return 'aggs' in es_query
+
+IGNORE_KEYS = [
+    'size',
+    'scroll',
+    'from',
+    'aggs',
+]
```
JeltevanBoheemen, line 33: Move constant to the top of the file

backend/api/tests/test_save_query.py
```diff
@@ -0,0 +1,45 @@
...
+def test_recent_queries(auth_user, saved_query):
+    assert saved_query in recent_queries(auth_user)
+
+    saved_query.started = timezone.datetime(2000, 1, 1, 0, 0, tzinfo=timezone.get_current_timezone())
+    saved_query.completed = timezone.datetime(2000, 1, 1, 0, 0, tzinfo=timezone.get_current_timezone())
+    saved_query.save()
+
+    assert saved_query not in recent_queries(auth_user)
+
+def test_same_query():
```
JeltevanBoheemen, line 33: Nice test!

backend/es/views.py
```diff
@@ -62,4 +68,22 @@ def post(self, request, *args, **kwargs):
...
+    def _log_query_started(self, request, corpus_name, es_query):
+        if save_query(request.user, es_query):
+            corpus = Corpus.objects.get(name=corpus_name)
+            return Query.objects.create(
+                user=request.user,
+                corpus=corpus,
+                query_json=es_query,
+            )
+
+    def _log_query_done(self, query, results):
```
JeltevanBoheemen, line 85: These functions don't log, but rather create/save, would change the name to indicate this.

JeltevanBoheemen (approved):
Good job. Suggest a few renames of functions.

Issue #1102, opened 2023-04-19
Save favourite corpora


lukavdplas:
A feature that would be nice if we add significantly more corpora: let users mark corpora as their favourites. A user's favourite corpora should appear at the top of the home page / corpus selection menu when they open it.

Adds some convenience for navigating to the corpora that you use regularly.

PR #1325, opened 2023-11-16
Feature/corpora link in menu

### LINKED ISSUE #994: Show and filter corpus metadata in index
As we add more corpora, the current corpus overview will likely become cumbersome.

- [x] Update to a more compact view that displays more metadata of the corpus. For example, we could show the time period and the language(s), and the type of data (newspaper, funerary inscriptions, etc.) -> resolved by #1142
- [x] Add the option to filter the metadata. -> closed by #1142 
- [ ] The corpus dropdown in the navbar at the top should just be a link to the index, or replaced with a dropdown of bookmarked corpora (c.f. #1102 )

lukavdplas:
close #994 

This addresses the final point of that issue: that if there are a lot of corpora, a dropdown with corpora becomes impractical. Replaces the corpus dropdown with a link.

### DIFF
--- a/frontend/src/app/menu/menu.component.html
+++ b/frontend/src/app/menu/menu.component.html
@@ -1,7 +1,10 @@
 <nav class="navbar" *ngIf="currentUser" aria-label="primary navigation">
     <div class="container is-fluid">
         <div class="navbar-brand">
-            <a class="navbar-item" [routerLink]="['/home']"> <img src="/assets/logo.png" width="50" height="50"> I-ANALYZER</a>
+            <a class="navbar-item" [routerLink]="['/home']">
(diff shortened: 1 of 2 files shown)


lukavdplas:
Technically, the number of clicks stays the same (2), but there is some extra scrolling in this version.

A searchable dropdown is a bit of a trade-off because a search function is basically useless for new users, but I can imagine seasoned users like having a quick option to switch corpora.

In the future, I'd like to add an option to bookmark corpora (#1102) and provide quick access to bookmarked corpora from the navbar, but that's beyond the scope of the current project.

BeritJanssen (approved):

Issue #794, opened 2022-07-20
Keyboard interaction


lukavdplas:
#777 fixes the most major issues for keyboard users, but a few issues remain. Most of these are more minor and/or would require more work to fix.

- [x] The dropdown menus from the top bar can be opened and used, but their tabindex is appended to the end of the page. So once you open the corpus dropdown, you have to tab through the _entire_ page to get to the list of corpora. My proposed fix would be to use the standard bulma dropdown menu instead of the primeng menu, that seems to do it. You would need to edit the CSS to keep the menu's signature look.
- [x] When tabbing through the filters: if you focus on the date filter it immediately opens the menu, which is fine. However, if you leave the menu by tabbing to the next control or pressing escape, the date filter is set to "active" even though you did nothing. (closed by #1139)
- [ ] The tab order of the pagination controls does not match left-to-right order, which is unexpected but not unusable.
- [x] Several buttons are `<a>` elements when they should be `<button>`, or [explicitly add keyboard interaction](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role). This applies to:
	- [x] The download table data button for visualisations
	- [x] The "show all fields" button in the fields dropdowns for searching and download
	- [x] The descending/ascending toggle button for sorting results
	- [x] The buttons at the footer of the document view (closed by #1169)
	- [x] The search results / visualisions tabs (closed by #1168)
- [x] When opening a document, tabbing continues to cycle through the controls in the background. Interaction with the background should be disabled while viewing a document, and pressing tab should focus on the first control _within_ the document view. (closed by #1169 )
- [ ] Charts do not have keyboard interaction. Mouse interactions are: tooltips showing the values of datapoints, zooming in the barcharts, dragging nodes in the nearest neighbours network. None of these show data that cannot be accessed through the table view or filtering, but those are more cumbersome routes.
- [x] The range filter triggers an update with the `onSlideEnd` event of the slider, which is only triggered when using a mouse to drag the slider. It is not triggered when using the keyboard (or clicking on the slider, for that matter)
- [x] In the search history, each query links to the relevant search request, but these links are not accessible by keyboard.
- [x] ~In the corpus header, the corpus information button has no indication that it receives focus.~ (obsolete since #1143)
[...]

PR #1327, opened 2023-11-20
Navbar improvements

### LINKED ISSUE #989: Show interface before login
Currently the first view when navigating to I-analyzer is the login screen. It would be nice if the home page showed more of the actual interface, with the login menu on the side or such.

lukavdplas:
Some accessibility improvements to the navbar:
- The menu toggle on mobile is now keyboard accessible
- The user dropdown has improved keyboard accessibility (see #794)
- Adds "skip to content link" at the start
- With the exception of logging out, menu items have an `href` link, rather than triggering a script event that sets the route. This makes links more predictable for users.
- Use `<header>` and `<footer>` in the app component so the header/footer get appropriate roles.

I've also cleaned up the logic of the navbar CSS and typescript. There were some arcane bits of logic without an adequate explanation of what they were for, so I *hope* it all works as before.

One final improvement is that the header is now visible when logged out, which also means new users can view the "about" page before registering. (close https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/989 )

### DIFF
--- a/frontend/src/app/menu/menu.component.html
+++ b/frontend/src/app/menu/menu.component.html
@@ -1,5 +1,9 @@
 <nav class="navbar" *ngIf="currentUser" aria-label="primary navigation">
     <div class="container is-fluid">
+        <a class="skip-to-content-link"
+            [routerLink]="['.']" fragment="main">
+            Skip to content
+        </a>
         <div class="navbar-brand">
             <a class="navbar-item" [routerLink]="['/home']">
                 <img src="/assets/logo.png" aria-hidden="true" width="50" height="50">
--- a/frontend/src/app/menu/menu.component.html
+++ b/frontend/src/app/menu/menu.component.html
@@ -116,7 +116,7 @@
                     </a>
                     <a class="navbar-item"
                     [routerLink]="['/registration']" [routerLinkActive]="'is-active'">
-                        Sign up
+                        Register
                 </a>
                 </ng-template>
             </div>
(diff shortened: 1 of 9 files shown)


frontend/src/app/menu/menu.component.html
```diff
@@ -1,50 +1,113 @@
...
-                    <span aria-hidden="true"></span>
-                    <span aria-hidden="true"></span>
+            <a  role="button" tabindex="0"
+                (click)="toggleMenu($event)" (keydown.enter)="toggleMenu($event)"
+                class="navbar-burger" [class.is-active]="menuOpen"
+                aria-label="menu" [attr.aria-expanded]="menuOpen"
+                data-target="navigation-menu">
+                <span aria-hidden="true"></span>
+                <span aria-hidden="true"></span>
+                <span aria-hidden="true"></span>
```
JeltevanBoheemen, line 23: Why the repeated span?

frontend/src/app/menu/menu.component.ts
```diff
@@ -1,118 +1,130 @@
...
+
+    toggleDropdown() {
+        this.dropdownOpen$.next(!this.dropdownOpen$.value);
     }
 
-    public gotoAdmin() {
-        window.location.href = environment.adminUrl;
+    toggleMenu() {
+        this.menuOpen = !this.menuOpen;
     }
```
JeltevanBoheemen, line 77: Inconsistent approach between these two methods.

JeltevanBoheemen (changes requested):
Code looks good and makes sense. 

One important request: I like the visible header when not logged in/registered. But the way back to the login page from the about page is not apparent. Add a nav element to that effect.

Laughed at the description of 'arcane' code, I noticed this when working on the auth service. It all seems to work. Reminder to thoroughly test solis/regular logins AND registration/email confirmation.

frontend/src/app/menu/menu.component.html
lukavdplas: I thought this was a mistake, but this is how the bulma documentation does it. My best guess is that this is how they create the "hamburger icon". I haven't tried removing it though.

PR #1331, opened 2023-11-23
deterministic order for documents

### LINKED ISSUE #1328: Failing backend test
[This build action failed](https://github.com/UUDigitalHumanitieslab/I-analyzer/actions/runs/6932088339), [and so did this one](https://github.com/UUDigitalHumanitieslab/I-analyzer/actions/runs/6943242575/job/18887878064).

One of the unit tests for the peace portal corpora fails. Note that these are checks for feature branches, but ones that only make changes to the frontend.

@BeritJanssen , can you help out with this?

lukavdplas:
close #1328 

My educated guess guess is that the errors from peace portal tests are caused by `os.walk` being non-deterministic on some operating systems. This fixes that.

### DIFF
--- a/backend/corpora/peaceportal/tests/test_peace.py
+++ b/backend/corpora/peaceportal/tests/test_peace.py
@@ -252,8 +252,10 @@ def test_imports(peace_test_settings, corpus_object):
     resulted_fields = set()
 
     docs = get_documents(corpus, start, end)
+    sorted_docs = (doc for doc in sorted(docs, key=lambda doc: doc['id']))
+
     for target in corpus_object.get('docs'):
-        doc = next(docs)
+        doc = next(sorted_docs)
         for key in target:
             tested_fields.add(key)
             assert key in doc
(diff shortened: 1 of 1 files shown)


JeltevanBoheemen (approved):

BeritJanssen:
Since the problem persists, I may instead go into the corpus definitions and see if I can use `glob` over `os.walk`. I also assume it has something to do with non-deterministic behaviour of `os.walk`. `glob` seems to be more stable overall.

lukavdplas:
I know the behaviour of `os.walk` is indeed not deterministic on all operating systems. The [documentation of glob](https://docs.python.org/3/library/glob.html) doesn't make any mention of this, other than saying the order is arbitrary.

JeltevanBoheemen:
I may have missed a part of this discussion, but shouldn't sorting the documents resolve this regardless of (non-)deterministic behaviour of walk or glob?
In other words, is there another cause for the failure?

lukavdplas:
Possibly. I had initially assumed this was the problem, but perhaps something else is going on.

BeritJanssen:
The code in the Peaceportal base class does a depth-first search for xml documents, yielding them at the leaves. Sorting the leaves then doesn't have an effect. In fact, the sorting was already in place, but it should be directory names that should be sorted for this to work.
```
for directory, _, filenames in os.walk(self.data_directory):
            for filename in sorted(filenames):
```
`glob` returns all paths (not just filenames) that conform to the pattern into a flat list, which will have a predictable order after sorting.

Issue #1324, opened 2023-11-16
State reset for document tagging controls


lukavdplas:
### What went wrong?

The state of the document tagging controls does not get reset when closing a document and moving to another. For instance, if you were in the middle of typing a new tag, it will open with the text input.

### What did you expect to happen?

When switching documents, the state of the control should be completely reset.

### Where did you find the bug?

- a local server

### Version

develop

### Steps to reproduce

- Open a document modal from the search results
- Select a tag from the dropdown, or start typing in a new tag
- Close the document and open another

Issue #1321, opened 2023-11-16
Overflow of long tags


lukavdplas:
### What went wrong?

Very long tags are displayed in an impractical manner: they can grow out of their container and overlap with the text of the document.

### What did you expect to happen?

Tags could be truncated or have text wrapping.

### Screenshot

![screenshot showing a long tag obscuring the text of the document](https://github.com/UUDigitalHumanitieslab/I-analyzer/assets/43678097/7569466f-0d6b-4db2-bd8b-889ed7d378b8)

### Where did you find the bug?

- a local server

### Version

develop

### Steps to reproduce

- Create a long tag for a document.
- Look at the document view (in the popup or its own page).

PR #1332, opened 2023-11-23
Bugfix/reset tagging controls

lukavdplas:
Resolves two bugs for the tagging controls, #1321  and #1324


JeltevanBoheemen (approved):
Good finds.

PR #1334, opened 2023-11-27
enable click and keyboard controls for slider

lukavdplas:
fixes a point from #794 : the range filter could only be controlled by dragging. You can now also trigger updates when controlling it with a keyboard or clicking it (without dragging).

Changes to the value are debounced by 100 ms to prevent constant updates. That's not really noticeable, but it does mean the slider can update while the user is still dragging. I'm not really sure if this would cause annoyance on larger corpora; it could be fixed with a more complicated solution.


JeltevanBoheemen (approved):
Implementation is fine. 

Concerning updates triggering on larger corpora being annoying, maybe a slightly higher debounce?
And did you try the [onSlideEnd](https://primeng.org/slider#api.slider.emitters.onSlideEnd) event over `onChange`?

lukavdplas:
onSlideEnd is what we did before, that that only triggers for drag events, which was the issue.

JeltevanBoheemen:
Right.

Issue #976, opened 2022-11-25
Add option to delete personal data


lukavdplas:
Users should have an option to have their personal data purged from the database.

Concretely, this means:
- deleting your search history
- deleting your downloads
- deleting your account (including search history and downloads)

JeltevanBoheemen:
- [ ] search history
- [ ] downloads
- [ ] account

lukavdplas:
Re. downloads, I remember why said this may not be necessary.

In https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/947, I suggest that downloads should have an "expiration date". It makes sense to me that we just delete downloads after, say, 90 days. You could still add an option to delete all your saved downloads right now, but it's less necessary in that case.

lukavdplas:
We recently added deleting accounts to lettercraft: https://github.com/CentreForDigitalHumanities/lettercraft/pull/157

Since lettercraft's user system is based on I-analyzer, it should be easy enough to take this PR and adapt it back to I-analyzer.

PR #1336, opened 2023-11-28
Feature/delete downloads

JeltevanBoheemen:
This partially implements #976, only the download part. Even with an expiration period, it's nice to offer a direct way. Creating a PR now because deleting accounts is slightly more involved, and may take some time to get around to for me.

Tried really hard to fit the `delete all` button underneath the `Delete` header, but couldn't get PrimeNG to comply.


frontend/src/app/history/download-history/download-history.component.html
```diff
@@ -7,6 +7,12 @@ <h1 class="title">Downloads</h1>
                 Downloads
             </div>
         </ng-template>
+        <ng-template *ngIf="downloads.length" pTemplate="summary">
+            <button class="button is-primary" type="reset" (click)="deleteAllDownloads()">
+                <span class="icon"><fa-icon [icon]="faTrash"></fa-icon></span>
+                <span>Delete all downloads</span>
+            </button>
+        </ng-template>
```
lukavdplas, line 15: I agree this is a good place for a delete button, but it's inconsistent that downloads are deleted here, while queries are deleted from the settings menu. The pages are almost identical, so they should handle this the same way. Perhaps you could move the "delete search history" button?

frontend/src/app/history/download-history/download-history.component.html
```diff
@@ -48,6 +55,16 @@ <h1 class="title">Downloads</h1>
...
+                <td>
+                    <a *ngIf="download.status === 'done' || download.status ==='error'" role="button" tabindex="0"
+                        (click)="deleteDownload(download)"
+                        (keydown.enter)="deleteDownload(download)">
+                        <span class="icon">
+                            <fa-icon [icon]="faTrash"></fa-icon>
+                        </span>
+                        <span>delete download</span>
+                    </a>
+                </td>
```
lukavdplas, line 67: Nice.

Addition my earlier comment: if you choose to make the search history menu more like this page, don't bother adding this action. The search history includes a unique filter, which means deleting one query would usually cause another identical one to appear. A decent delete button per query would take more time.

lukavdplas (changes requested):
Nice, very neat! :)

I have one comment re. the interface in the downloads and search history page.

JeltevanBoheemen:
Looking at this later!

PR #1338, opened 2023-11-29
bugfix/search history template

### LINKED ISSUE #1337: Search history broken
### What went wrong?

The search history page did not load. Error message in console:

```
Uncaught (in promise): TypeError: right-hand side of 'in' should be an object, got undefined
```

The page should have loaded.

### Screenshot

screenshot showing empty page

- https://ianalyzer.hum.uu.nl
- https://peopleandparliament.hum.uu.nl

### Version

5.2.0

### Steps to reproduce

After logging in, go to `/search-history`

I've seen this on multiple servers, but I couldn't reproduce it locally, even when switching to the 5.2.0 version. Not sure what causes the difference.

lukavdplas:
this seems to resolve #1337 on the test server 😅

### DIFF
--- a/frontend/src/app/history/search-history/query-filters.component.ts
+++ b/frontend/src/app/history/search-history/query-filters.component.ts
@@ -16,7 +16,7 @@ export class QueryFiltersComponent implements OnInit {
     constructor() { }
 
     ngOnInit() {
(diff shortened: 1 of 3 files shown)


BeritJanssen (approved):
Wonderful!

PR #1356, opened 2023-12-14
Resolve all frontend linting issues

JeltevanBoheemen:
In a frustrated mood I decided to kill some linting errors and warnings. 
There is a lot of auto formatting, so a full review might be impractical. I advise you to look at `.eslintrc.json` and see if you agree with the rule changes. Removed rules did not generate any linting warnings/errors, so were deemed superfluous.

PR #1357, opened 2023-12-14
add date_range mapping type

### LINKED ISSUE #1292: standardised nested date mapping
Proposed solution for various issues surrounding date fields, including:
- Mixed precision levels which would complicate https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/992
- https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1114
- The awkward UI of having four (!) date pickers for "earliest earliest date", "latest earliest date", etc.

I would propose using a single nested field mapping for all date fields. For instance, we can use four subfields:

- `exact`: the exact day, if known
- `year`: the exact year, if known
- `lower`: lower bound for the date
- `upper`: upper bound for the date

`exact` and `year` would not always be filled in, of course, but you could make it so the "less precise" values are inferred. (Either during data entry or by elasticsearch.) So in a corpus with exact dates, you could still request `date.year`, `date.lower` and `date.upper`.

Thus, a date filter for 1800-1900 would look something like the following:

```json
{
	"range": {
		"date.lower": {
			"gte": "1800-01-01",
			"format": "yyyy-MM-dd"
		},
		"date.upper": {
			"lte": "1900-01-01",
			"format": "yyyy-MM-dd",
		}
	}
}
```

Which should work on any corpus with a `date` field, regardless of precision.

There is a broader question whether standardising corpus fields is a good idea (c.f. https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/993), but then date fields are already "lightly" standardised by being referenced in the corpus metadata.

BeritJanssen:
Close #1292 .


lukavdplas (commented):
Looks good!

Can you add a screenshot of what this looks like in the document view?

lukavdplas:
By the way, could you adjust your autoformatter to only format your changes rather than the whole file? If you want to format existing code, I would prefer a separate PR like #1356

BeritJanssen:
This is how it now looks in the document overview for a testset from the peaceportal-epidat corpus:
Screenshot 2023-12-20 at 14 53 33

PR #1361, opened 2023-12-15
Unify icon library

lukavdplas:
We were currently using both font awesome 4 and font awesome 6 icons. This updates all remaining FA4 icons to FA6.

Also makes a few minor template fixes while I was going through the icons - mostly adding `aria-label` or `aria-hidden` to icons.


frontend/src/app/filter/filter-manager.component.ts
```diff
@@ -15,6 +16,11 @@ import { FilterInterface, QueryModel } from '../models/index';
 export class FilterManagerComponent {
     @Input() queryModel: QueryModel;
 
+    icons = {
+        clear: faTrashCan,
+        toggle: faFilter,
+    };
+
```
BeritJanssen, line 23: can we factor this out? This is now repeated here and in `filter-box-component`.

BeritJanssen (changes requested):
🚀  Nice work! I do wonder though whether we could factor out all the 
```javascript
icons = {
   name: iconName
}
```
objects into one `icons` utility function "to rule them all", that components which need icons can import. This would be DRYer. Or are the performance concerns with that approach?

JeltevanBoheemen (approved):
I can get behind this approach!

PR #1385, opened 2024-01-16
Hotfix/5.4.2

### LINKED ISSUE #1384: Long download mail breaks
### What went wrong?

The email for long downloads checks for a variable named "login" which is missing in the context. This prevents the email from sending.

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/7170f219453f9cee71112f5b3b9ad589ecba9f93/backend/download/templates/download_mail.html#L117

Only seems to affect downloads over 10.000 results (for users with that permission).

As a side note, this happens after the creating of the CSV file, so it will still be available for download, but you can't download it from the interface either because the status is set to "error". The error handling for emails should not affect the status of the downloaded file. That can be left to a future issue, though.

Email should send.

### Screenshot

_No response_

- https://ianalyzer.hum.uu.nl

### Version

5.4.0

### Steps to reproduce

- Log in with an account that can download more than 10.000 documents
- Make a search request with over 10.000 results
- Use the "download CSV" button to download the results
- Wait for the download to complete - the status will change to "error" in your downloads page.
- Check the celery log on the server for the error traceback

lukavdplas:
close #1384

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -35,5 +35,5 @@ keywords:
   - elasticsearch
   - natural language processing
 license: MIT
-version: 5.3.0
-date-released: '2023-12-04'
+version: 5.4.2
+date-released: '2024-01-16'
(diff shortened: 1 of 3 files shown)


JeltevanBoheemen (approved):

jgonggrijp:
Was this hotfix abandoned or merged? In the latter case, why was it necessary to close it manually?

lukavdplas:
I must have closed it before using git-flow to handle merging, which wasn't necessary.

PR #1387, opened 2024-01-17
Feature/cancel tasks

### LINKED ISSUE #895: Allow cancelling/preventing ngram loading
The ngram graph has a lot of options and takes a while to load, so it can be annoying to experienced users (and developers) that already know they want to pick options other than the default.

> Ah, and while I think about it: would it be possible to set the parameters for the visualisation (once again, esp ngrams) before it calculates it ? Right now, I have to wait two times: for the tool to load with parameters I don’t care about (the default) and a second time for those I want. Or could this be an option we enable depending of the user’s preferences ? (If I am just exploring the data, may be it makes senses that somethings launches automatically)

We could of course just not load the graph on intialisation (and wait for the user to confirm it) but this may be confusing to users seeing it for the first time. There is just a big menu with options and they have no idea what the end result would look like.

Perhaps a way to satisfy all users would be to allow them to abort the loading (which should also send the cancel task request to the backend).

BeritJanssen:
This PR is meant to fix #895 . The task polling is not cast to Promise anymore, and be completed when calling `.next()` on a `stopPolling` Subject.


frontend/src/app/visualization/barchart/barchart.directive.ts
```diff
@@ -372,19 +374,27 @@ export abstract class BarchartDirective
...
         const queryModelCopy = this.queryModelForSeries(series, queryModel);
-        return this.requestSeriesTermFrequency(series, queryModelCopy).then(result =>
-            this.apiService.pollTasks<TermFrequencyResult>(result.task_ids)
-        ).then(res =>
-            this.processSeriesTermFrequency(res, series)
-        ).catch(error => {
-            console.error(error);
-            this.error.emit(`could not load results: ${error.message}`);
-            return series;
+        this.requestSeriesTermFrequency(series, queryModelCopy).then(response => {
```
lukavdplas, line 379: This was already the case, but the logic in this function shouldn't really be handled in the component.

Ideally, the component should just use `this.visualisationService.getTheData(this.destroy$).then(...)` and not care whether the service is fetching the data through polling or in a single request.

But like I said, that was already an issue, so feel free to ignore.

lukavdplas (changes requested):

BeritJanssen:
All good points, but as indicated by the Draft status, this is not ready for review yet.

lukavdplas:
Oh, my bad!

frontend/src/app/visualization/barchart/barchart.directive.ts
```diff
@@ -372,19 +374,27 @@ export abstract class BarchartDirective
...
+            this.tasksToCancel = response.task_ids;
+            const poller$ = this.apiService.pollTasks<TermFrequencyResult>(this.tasksToCancel);
+            poller$.subscribe({
+                error: (error) => this.onFailure(error),
+                next: (result) => {
+                    this.apiService.stopPolling$.next();
+                    this.processSeriesTermFrequency(result['results'], series);
+                }
+            });
```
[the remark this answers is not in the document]
BeritJanssen: Yes, so would I. But that would mean overhauling the visualization service's `aggregateTermFrequencySearch` to return an Observable. I think it's clearer now than it was before, where api.service internally converted the Observable to a Promise.

frontend/src/app/visualization/barchart/barchart.directive.ts
```diff
@@ -372,19 +374,27 @@ export abstract class BarchartDirective
...
-            console.error(error);
-            this.error.emit(`could not load results: ${error.message}`);
-            return series;
+        this.requestSeriesTermFrequency(series, queryModelCopy).then(response => {
+            this.tasksToCancel = response.task_ids;
+            const poller$ = this.apiService.pollTasks<TermFrequencyResult>(this.tasksToCancel);
+            poller$.subscribe({
+                error: (error) => this.onFailure(error),
+                next: (result) => {
+                    this.apiService.stopPolling$.next();
```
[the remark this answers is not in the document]
BeritJanssen: A shared subject is probably going to cause bugs indeed. Every component passes its own `stopPolling$` subject now.

frontend/src/app/visualization/barchart/barchart.directive.ts
BeritJanssen: I'm not sure I understand. The current setup (once we get to this line) is that we first initiate the tasks, and then, with the task ids, start polling. If we are in this function (`getTermFrequencies`), we are always polling.

frontend/src/app/services/api.service.ts
```diff
@@ -108,42 +109,32 @@ export class ApiService {
...
         return this.http
-            .post<TaskSuccess>('/api/task_status', data)
+            .post<TaskSuccess>('/api/abort_tasks', data)
             .toPromise();
     }
 
-    private tasksDone<ExpectedResult>(response: TasksOutcome<ExpectedResult>) {
-        return response.status !== 'working';
+    private tasksDone(response: TasksOutcome) {
+        return response.status === 'done';
```
lukavdplas, line 126: how will a failed task stop polling?

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -289,6 +290,7 @@ export class NgramComponent extends ParamDirective implements OnChanges {
     }
 
     confirmChanges() {
+        // this.isLoading.emit(true);
```
lukavdplas, line 293:
```suggestion
```

BeritJanssen:
I think the setup is reasonable now. I lost a lot of time trying to fix the test - unfortunately, it's not possible to access Karma from the current Docker setup, so I cannot set any breakpoints. What seems to happen is that the mock data in `api.ts` for the ngram.component test returns `{status: 'done', results: [[fakeNgramResult]]}`, instead of `[fakeNgramResult]`, as I would expect. Without being able to set breakpoints it's terribly annoying to understand why data is being wrapped in an array here.

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -280,6 +279,8 @@ export class NgramComponent extends ParamDirective implements OnChanges {
         }
 
         this.parametersChanged = true;
+        this.stopPolling$.next();
+        this.isLoading.emit(false);
```
lukavdplas, line 283: This is behind on develop.

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -196,51 +200,46 @@ export class NgramComponent extends ParamDirective implements OnChanges {
...
+    onFailure(error: {message: string}) {
+        this.stopPolling$.next();
+        console.error(error);
         this.currentResults = undefined;
         this.ngramError.emit(error.message);
         this.isLoading.emit(false);
     }
 
     onDataLoaded(result: NgramResults) {
+        this.stopPolling$.next();
```
lukavdplas, line 235: This is a bit confusing because I would think this method is triggered because the poller has finished.

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -196,51 +200,46 @@ export class NgramComponent extends ParamDirective implements OnChanges {
...
+                this.currentParameters).then(
+                    response => {
+                        this.tasksToCancel = response.task_ids;
+                        // tasksToCancel contains ids of the parent task and its subtasks
+                        // we are only interested in the outcome of the parent task (first in array)
+                        const poller$ = this.apiService.pollTasks([this.tasksToCancel[0]], this.stopPolling$);
+                        poller$.subscribe({
+                            error: (error) => this.onFailure(error),
+                            next: (result: SuccessfulTask<NgramResults[]>) => this.onDataLoaded((result).results[0]),
+                            complete: () => this.apiService.abortTasks({ task_ids: this.tasksToCancel })
```
lukavdplas, line 220: It looks odd to cancel tasks that have seemingly just completed.

lukavdplas (commented):
Okay, for realskies review.

Some of the logic looks a bit sketchy to me, as pointed out in my comments - which isn't to say it's wrong, just that it could do with some unit tests and/or explanatory comments.

General (non-essential) advice:
- I think it would be clearer if you split the `stopPolling$` observable into the individual event streams that you care about, e.g. `parametersChanged$`, `destroy$`, and merge those into a `stopPolling$` observable. Right now, it is easy to see what the intended callback for `stopPolling$` is, but it's difficult to read the logic of the event stream it represents.
- The point I made earlier is that it's often clearer if components are more focused on their job as views / controllers, and don't have detailed logic to manage API interactions. Essentially, the component should write `result$ = service.fetchDataIWant(parameters, cancel$)` and whether that method is (internally) polling or making a direct request isn't the component's business. But this is a more general point of making flexible components. As I said, this already isn't the case right now, so you don't need to address it here.

frontend/src/app/services/api.service.ts
BeritJanssen: It throws a hard error now from the backend.

frontend/src/app/visualization/ngram/ngram.component.ts
BeritJanssen: This line indeed means we fire off a request to abort a completed task. Slightly inefficient, but it does avoid having to call `abortTasks` everywhere in tandem with stopPolling. Perhaps the complete handler could still check the type of result though to avoid this.

frontend/src/app/visualization/ngram/ngram.component.ts
BeritJanssen: No, the result is in, but the poller doesn't stop polling without this.

backend/api/views.py
```diff
@@ -60,11 +63,6 @@ def post(self, request, *args, **kwargs):
         if all(result.state in ['PENDING', 'STARTED', 'SUCCESS'] for result in results):
             return Response({'status': 'working'})
 
-        # some tasks failed
-        for result in results:
-            logger.error(result.info)
```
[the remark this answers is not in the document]
BeritJanssen: The log files that are interesting here are in the celery logs, I would think.

BeritJanssen:
`result$ = service.fetchDataIWant(parameters, cancel$)` - I like this idea, but in practice, the fact that ngrams have a finalizing task after the parallel tasks have run, and that term frequencies return as a array of the parallel tasks themselves, is actually going to make this hard to implement.

frontend/src/app/visualization/ngram/ngram.component.ts
lukavdplas: Ah right, I see you removed that logic. Why?

BeritJanssen: Which logic do you refer to? You mean the `take(1)` on `api.service.pollTasks`? With this I reasoned that it would be cleaner to just have the one `takeUntil` operator in the chain.

BeritJanssen: Then again, the `take(1)` operator actually means we don't have to call `stopPolling$.next()` for success/ failure in two components. I think I'll reintroduce it.

PR #1390, opened 2024-01-17
Feature/angular lifecycle fix

### LINKED ISSUE #1365: Angular lifecycle conflict
Angular linting complains about lifecycle conflict: in `visualization/visualization.component.ts`

```
   27:16  error  Implementing DoCheck and OnChanges in a class is not recommended          @angular-eslint/no-conflicting-lifecycle
   27:25  error  Implementing DoCheck and OnChanges in a class is not recommended          @angular-eslint/no-conflicting-lifecycle
  115:5   error  Declaring ngDoCheck and ngOnChanges method in a class is not recommended  @angular-eslint/no-conflicting-lifecycle
  121:5   error  Declaring ngDoCheck and ngOnChanges method in a class is not recommended  @angular-eslint/no-conflicting-lifecycle
```

Resolve these (justified) errors. Should implement with either `DoCheck` or `OnChanges`, not both.

lukavdplas:
Fixes the NgDoCheck / NgOnChanges lifecycle conflict in the visualisation component (close #1365 )

This is done the way I suggest in that issue, by letting visualisations handle their own loading spinner rather than setting it from the parent component. For consistency, this is also implemented for the word model component.

### DIFF
--- a/frontend/src/app/visualization/barchart/barchart.directive.ts
+++ b/frontend/src/app/visualization/barchart/barchart.directive.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/member-ordering */
-import { Directive, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges } from '@angular/core';
+import { Directive, EventEmitter, Host, HostBinding, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges } from '@angular/core';
 
 import * as _ from 'lodash';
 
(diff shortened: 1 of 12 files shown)


BeritJanssen (approved):
Nice that this is fixed! Could the `@HostBinding` for the loading also be part of a Directive to inherit from? Just an idea as now code is repeated over multiple components -- but then, there is certainly not more repetition than before.

lukavdplas:
i tried that, but it's tricky when the component is already inheriting the ParamDirective.

PR #1399, opened 2024-01-24
use Docker for two separate workflows

BeritJanssen:
The branch splits the unit tests up in a frontend / backend component, so that they can be run in parallel. It was easiest to set this up using Docker. The images built for this purpose will be cached, so it should be pretty fast.


JeltevanBoheemen (approved):
That's awesome! We should change the required checks when this is merged:
image

PR #1402, opened 2024-01-26
lazy load document tags

### LINKED ISSUE #1319: Tags: only request document tags when needed
The logic of fetching document tags in https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/1199 was something I intended to change later, but then forgot about.

Right now, document tags are fetched whenever a `FoundDocument` is created in the frontend, but this should only happen when the tags are actually used.

In practice, tags are only visible when you open a document, so when you load a page of search results, you don't need to load the tags for all 20 documents on the page. Right now, though, those tags are requested regardless, so every page makes 20 requests to the backend to get the tags for each document.

This _could_ be resolved easily by just defining `document.tags$` as an observable with `shareReplay(1)`, but then you have the issue that you need to update the value of `tags$` when users add and remove documents.

lukavdplas:
close #1319


frontend/src/app/models/found-document.ts
```diff
@@ -37,7 +39,13 @@ export class FoundDocument {
         this.relevance = hit._score / maxScore;
         this.fieldValues = Object.assign({ id: hit._id }, hit._source);
         this.highlight = hit.highlight;
-        this.fetchTags();
+
+        const created$ = timer();
```
JeltevanBoheemen, line 43: Could you explain the use of this `timer`?

frontend/src/app/models/found-document.ts
```diff
@@ -74,24 +82,28 @@ export class FoundDocument {
...
     }
 
     removeTag(tag: Tag): void {
-        const newTags = _.without(this.tags$.value, tag);
-        this.setTags(newTags);
+        this.tags$.pipe(
+            take(1),
+            map(tags => _.without(tags, tag)),
+            mergeMap(tags => this.setTags(tags)),
+            tap(() => this.tagsChanged$.next()),
```
JeltevanBoheemen, line 98: As a personal preference I would put this in the `subscribe` block, avoid another `pipe` part

Issue #1403, opened 2024-01-26
Refactor routing / routed state management


lukavdplas:
This issue is intended to track something we discussed in our last meeting and that I'm currently working on.

Deep routing for queries is a bit of a mess. The query parameters of the route are supposed to maintain two-way synchronisation with the state of the UI. There are currently two ways in which this is implemented (depending on the component):

- The `ParamDirective` is a superclass for components. The idea is that components handle state updates by sending updates to the `Router`, and then listen to changes in the route. This is a clean way to achieve two-way binding, but tying everything to components makes it difficult to model interdepencies between different parts of the query / results. It also makes it difficult to manipulate the data without using that routed component, which complicates unit testing.
- For the reasons described above, the `QueryModel`, as a more complex model, is not really built on the `ParamDirective` class, but the implementation is not very clean. The idea is that the model handle's its state internally, and the component that creates it (the `SearchComponent`) makes sure the route is kept up to date. If the route is changed in a way that doesn't match the current state of the model, the model is re-instantiated from the route. This carries a risk that if the conversion between the route and the model is incorrect, the application can get caught in a loop.

So the main issues are (in no particular order):
1. inconsistent implementation
2. difficulty in managing connected parameters
3. difficulty in unit testing
4. two-way binding that can cause cyclical updates

I've been working on a solution that should help with issues 2-4. If it is applied across the application, point 1 would be addressed as well.

lukavdplas:
Current status of this issue:

After https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/1467 is merged, the core of the application is based on the `Store` / `StoreSync` classes to handle routing.

The refactor will be complete if we also convert:
- the word models query
- visualisation modules

For visualisations, this can be done by [basing visualisations on the Results class](https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1468) but may also be done by rewriting the component so it inherits `StoreSync`. The former option is more "proper" but also takes longer.

jgonggrijp:
Perhaps this is a superfluous comment, but you could consider the route a special type of view/controller. As such, ideally it would communicate changes by editing a model and only update itself in response to changes in a model. This would eliminate any direct ties between routes and components and avoid cyclical updates. Maybe this what you were planning to do already, or similar to it.

lukavdplas:
[...]

PR #1404, opened 2024-01-29
Models for route synchronisation

lukavdplas:
This update is part of #1403. It implements the base classes for improved routing/state synchronisation. It only defines the classes, so none of the existing functionality is changed.

This system can be introduced gradually, but the intention is that _every_ query parameter is eventually managed through the classes that are introduced here. I'm almost finished implementing this on the QueryModel and PageResults model, which I'll put in separate PRs.

There is a documentation update included that gives a good conceptual overview of the system. The key ideas in this approach are:

- The class that manages the query parameters of the route (`RouterStore`) has a drop-in replacement that just uses an internal observable (`SimpleStore`). This makes it much easier to test things and use models in a more flexible way.
- Models can extend the `Stored` class to work in connection with a store. An independent model can be instantiated with  `new SimpleStore()`, while a routed model can use the `RouterStore` provided through dependency injection.
- Updates on a `Stored` model happen by sending the update to the store object, which the model observes.

*Requested level of review:* as I mentioned, I'm almost done implementing this for queries and search results, so I have already tested the classes in actual use. Don't worry about checking the code on a technical level - it's been tested more thoroughly than a "dry" review can. What I *do* want feedback on is

- whether the code makes sense intuitively
- clarity of the code
- the quality of documentation

### DIFF
--- a/documentation/Frontend-deep-routing-and-state-management.md
+++ b/documentation/Frontend-deep-routing-and-state-management.md
@@ -8,7 +8,7 @@ class MyModel {
 }
 ```
 
-A _model_ handles data, API calls, and the state of the user's workflow, and is kept separate from Angular components. Components acts as _views_ and/or _controllers_, and their script is solely aimed at translating between the UI and the model.
+A _model_ handles data, API calls, and the state of the user's workflow, and is kept separate from Angular components. Components act as _views_ and/or _controllers_, and their script is aimed at translating between the UI and the model.
 
 This keeps components lightweight and makes them more flexible: because the core data is implemented independently, you can split a component, swap a checkbox for a dropdown, etc. without messing up the data.
 
@@ -18,7 +18,7 @@ Usually, a model is created in the highest component that needs it, by calling `
 
 ### Dependency injection: providing services to models
 
-More complex models may require services to make API calls, create alerts, or listen to the global state of the application.
+More complex models may require services to make API calls, or listen to the global state of the application.
 
 In those cases, provide the service as a private member variable in the constructor of the model:
 
@@ -26,7 +26,6 @@ In those cases, provide the service as a private member variable in the construc
 class MyModel {
     constructor(
         private apiService: ApiService,
-        private notificationService: NotificationService,
     ) {
         // ...
     }
@@ -44,10 +43,9 @@ class MyComponent {
     model: MyModel;
 
     constructor(
-        private apiService: ApiService,
-        private notificationService: NotificationService,
+        private apiService: ApiService,,
     ) {
-        model = new MyModel(apiService, notificationService);
+        model = new MyModel(apiService);
     }
 }
 ```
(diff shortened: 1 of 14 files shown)


documentation/Frontend-deep-routing-and-state-management.md
```diff
@@ -0,0 +1,127 @@
...
+        return {
+            foo: state.foo
+        };
+    }
+}
+```
+
+The `Stored` class provides the main ways in which you can interact with the model:
+- a BehaviorSubject `state$` which tracks the latest state of the model based on the store.
+- a method `setParams()` which updates some or all of the properties in the model's state by sending an update to the store.
```
BeritJanssen, line 102: I don't see the `setParams` method here. Should this be `storeToState`?

frontend/src/app/store/router-store.ts
```diff
@@ -0,0 +1,41 @@
...
+    providedIn: 'root'
+})
+export class RouterStore implements Store {
+    params$: Observable<Params>;
+    paramUpdates$ = new Subject<Params>();
+
+    constructor(
+        private router: Router
+    ) {
+        this.params$ = this.router.routerState.root.queryParams;
```
BeritJanssen, line 20: Why do we need this Observable, if we already have a Subject + a `currentParams` function? Perhaps `paramUpdate$` as a `BehaviorSubject` with `this.router.routerState.root.queryParams` as initial value would be a more concise way to go?

frontend/src/app/store/simple-store.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    constructor() {
+        this.paramUpdates$.pipe(
+            scan(this.merge),
+            map(omitNullParameters),
+            map(obj => _.mapValues(obj, _.toString))
+        ).subscribe(params =>
+            this.params$.next(params)
+        );
+    }
```
BeritJanssen, line 25: Alright, here I understand the distinction between `paramUpdate$` and `param$`. Nice use of the `scan` function. Elegant! I'm not sure I agree with emulating the `router-store`'s behaviour here and converting everything to `string`. Should the router store not implement functions to infer & convert to `number` or `boolean` types instead?

lukavdplas: That sounds nice, but I don't think that can be done in a way where valid strings like `'5'` or `'true'` are preserved _and_ the routes remain backwards compatible.

frontend/src/app/store/stored.ts
```diff
@@ -0,0 +1,103 @@
...
+export abstract class Stored<State extends object> {
+    /**
+     * The stored state to of the model.
+     *
+     * This is implemented as a BehaviorSubject so you can read state$.value, but
+     * you should NOT push values onto the state directly. Use setParams() instead.
+     */
+    state$: BehaviorSubject<State>;
+
+    protected complete$ = new Subject<void>();
```
BeritJanssen, line 18: Upon first read of documentation and classes, I find it hardest to grasp what `Stored` is for, and what it does. Perhaps it would be clearer if you called this `StoreDirective`? Because this is what components will have to extend, right? I think adopting Angular-speak will help future developers.

BeritJanssen: Could we reimplement this as a directive, rather than a model?

frontend/src/app/store/simple-store.ts
BeritJanssen: I thought we'd always be able to convert back and fro, but of course user queries may include strings like '5' or 'true'. Hmm...

lukavdplas: The current solution that this conversion is left to the model implementation, which knows what type to expect. For instance, the `BooleanFilter` knows that `true` should be parsed as a boolean value, but for the query parameter, this would be interpreted as the string `'true'`.

PR #1405, opened 2024-01-29
Implement routing in PageResults

### LINKED ISSUE #1214: add pagination to route
When navigating results, add the page number to the route.

Some advantages:
- For longer sessions, it is nice if the page does not get reset when the page is refreshed.
- When users are going through a long list of results, they can bookmark the specific page they were at
- When opening a document's page, users can go back with browser navigation to the page they were at
- This would make it possible to open document context links at the page of the document.

Note: this should at least wait until #1162 is complete.

lukavdplas:
This update is part of #1403. It effectively does two things:

- Implemented the new routing models on the `PageResults` model.
- The `PageResults` model now manages the query parameters `sort`, `highlight`, and `page`. (The first two used to be handled by the `QueryModel`.)

I apologise for including both of these in the same PR. PageResults was the best candidate to work out cleaner state management because it was created with that mind - but it also needed to have a state to manage in order to do that.

If you prefer, you can check the changes up to 130e8d13107cd3508e40f2ae27aa4064e17e3796 separately; those handle the reallocation of reponsibility. Everything from f615590b02a1f3631969798b13a35402c8846e73 handles the implementation of route synchronisation in PageResults.

The reallocation makes the following changes:

- Responsibility for sorting and highlighting now lies with the `PageResults` model rather than the `QueryModel`
- Removes disabling logic  in the highlight selector component (the component was already excluded from the DOM when highlighting is disabled)
- Parameters for the `DownloadComponent` are now _more_ of a mess because it depends on the results tab which may or may not be active.
- Highlighting in downloads now uses the configured highlight size rather than a fixed value

The implementation of routing makes the following changes:

- `PageResults` is based on the `Stored` class.
- The current page is now included in the route. (close #1214)

This PR is a draft because it may need to change based on the review for #1404 - other than that, it's review-ready.


frontend/src/app/models/results.ts
```diff
@@ -25,24 +27,28 @@ export abstract class Results<Parameters, Result> {
...
+            this.setParams(params)
         );
 
-        this.result$ = this.parameters$.pipe(
+        const queryUpdate$ = merge(timer(0), this.query.update);
+
+        this.result$ = combineLatest([queryUpdate$, this.state$]).pipe(
             takeUntil(this.complete$),
+            map(_.last),
+            tap(() => this.error$.next(undefined)),
```
BeritJanssen, line 51: Doesn't this mean we throw an error for every Observable update?

frontend/src/app/models/results.ts
```diff
@@ -25,24 +27,28 @@ export abstract class Results<Parameters, Result> {
...
-            this.setParameters(params)
+            this.setParams(params)
         );
 
-        this.result$ = this.parameters$.pipe(
+        const queryUpdate$ = merge(timer(0), this.query.update);
+
+        this.result$ = combineLatest([queryUpdate$, this.state$]).pipe(
             takeUntil(this.complete$),
+            map(_.last),
```
BeritJanssen, line 50: Why is this necessary? It's not evident from type or similar that we'll be getting in arrays here.

frontend/src/app/models/results.ts
```diff
@@ -10,9 +12,9 @@ import * as _ from 'lodash';
  * for fetching results. Results will be loaded when the query model
  * or the parameters update
  */
-export abstract class Results<Parameters, Result> {
+export abstract class Results<Parameters extends object, Result> extends StoreSync<Parameters> {
```
BeritJanssen, line 15: Why does Parameters need to extend object?

frontend/src/app/utils/es-query.ts
```diff
@@ -10,6 +10,8 @@ import { findByName } from './utils';
 import { SearchFilter } from '../models/field-filter';
 import { APIQuery } from '../models/search-requests';
 import { TagFilter } from '../models/tag-filter';
+import { PageResultsParameters } from '../models/page-results';
+import { DeepPartial } from 'chart.js/dist/types/utils';
```
BeritJanssen, line 14: Is this import from `chart.js` correct?

frontend/src/mock-data/api.ts
```diff
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import { Subject, Observable, of } from 'rxjs';
 import { mockUserResponse } from './user';
+import { LimitedResultsDownloadParameters } from '../app/models';
```
BeritJanssen, line 4: Where exactly is this defined? Can you import from the specific model file for clarity? Given on the name of the type, I cannot quite fathom what this type describes.

BeritJanssen (changes requested):
Nice! I do have some points where documentation could be improved, or types / imports could clarify somewhat (see comments).

frontend/src/app/utils/es-query.ts
lukavdplas: It's not related to charts, but it's just a bit of laziness to import a `DeepPartial` typedef from there rather than define it ourselves.

frontend/src/app/models/results.ts
lukavdplas: No, it updates the stream of "errors thrown in the last result (if any)" to `undefined` - so it clears any existing errors

frontend/src/app/models/results.ts
lukavdplas: In this class, that's needed because the `State` type in `StoreSync` must be an object, so you can't allow `any` in the child class.

As for the `StoreSync`, it's needed to allow partial states as input to the `setParams` (i.e. only specify the new value for _some_ properties).

frontend/src/mock-data/api.ts
lukavdplas: I agree it would be clearer to remove `app/models/index.ts` (and some similar files) but that's a project configuration. I'll update the import here, in any case.

lukavdplas: It describes the parameters included in the body in a request for a download with a limited number of results (as opposed to the large download).

PR #1406, opened 2024-01-29
Improved routing in QueryModel

lukavdplas:
part of #1403 : this implements the models in #1404 on the `QueryModel` and filter classes.


frontend/src/app/models/base-filter.ts
```diff
@@ -112,27 +115,38 @@ export abstract class BaseFilter<InitialParameters, FilterData> implements Filte
...
-            const activate = !active;
-
-            if (activate) {
-                this.activate();
-            } else if (deactivate) {
-                this.deactivate();
-            } else if (active) {
-                this.update.next();
+
+            if (toDefault) {
```
BeritJanssen, line 122:
```suggestion
            if (this.isDefault(data) {
```
Not assigning this to a variable first is more readable IMO.

frontend/src/app/models/base-filter.ts
```diff
@@ -112,27 +115,38 @@ export abstract class BaseFilter<InitialParameters, FilterData> implements Filte
...
+            } else {
+                this.setParams({
+                    active: true,
+                    data,
+                });
             }
         }
     }
 
     /** reset the filter state: set the data to default an deactivate */
```
BeritJanssen, line 136: Comment string seems incomplete.

frontend/src/app/models/query.ts
```diff
@@ -73,35 +78,46 @@ export interface SearchParameters {
...
 
-export class QueryModel {
+export class QueryModel extends StoreSync<QueryState> {
     corpus: Corpus;
-	queryText: string;
-	searchFields: CorpusField[];
     filters: FilterInterface[];
 
-	update = new Subject<void>();
+	update: Observable<void>;
```
BeritJanssen, line 91: inconsistent indentation.

frontend/src/app/models/query.ts
```diff
@@ -128,46 +144,37 @@ export class QueryModel {
...
 
     /**
-     * convert the query to a parameter map
-     *
-     * All query-related params are explicity listed;
-     * empty parameters have value null.
+     * convert the query to a a parameter map, for deep linking
      *
+     * Unlike stateToStore(), this:
+     * - only includes explicited properties, it doesn't set null values
```
BeritJanssen, line 161: explicited -> explicit?

frontend/src/app/models/query.ts
```diff
@@ -204,9 +219,17 @@ export class QueryModel {
...
+        const keys = _.flatten([
+            this.keysInStore,
+            this.filters.map(filter => filter.keyInStore)
+        ]);
+
+        return this.store.params$.pipe(
+            takeUntil(this.complete$),
+            distinctUntilChanged(_.isEqual, params => _.pick(params, keys)),
+            skip(1),
+            map(to()),
```
BeritJanssen, line 232: Please document what these operators do, I find it hard to follow here.

frontend/src/app/search/search.component.ts
```diff
@@ -84,7 +85,10 @@ export class SearchComponent extends ParamDirective {
     async initialize(): Promise<void> {
         this.user = await this.authService.getCurrentUserPromise();
         this.corpusSubscription = this.corpusService.currentCorpus
-            .pipe(filter((corpus) => !!corpus))
+            .pipe(
+                filter((corpus) => !!corpus),
+                distinct(corpus => corpus.name),
```
BeritJanssen, line 90: Do we need the filter operator above? Shouldn't the `distinct` operator filter enough here?

frontend/src/app/utils/utils.ts
```diff
@@ -28,3 +28,9 @@ export const showLoading =<T>(isLoading: BehaviorSubject<boolean>, promise: Prom
         return result;
     });
 };
+
+/**
+ * returns a function that always returns `value`
+ */
+export const to = (value?: any) =>
+    () => value;
```
BeritJanssen, line 36: Why?

BeritJanssen (approved):
Nice! Some parts may benefit from some extra documentation. Ideally, we would disentangle the query text from the filters in the long run, and gather the state of the different UI components relating to queries once the user sends a query request to the backend. But that's something for after the backend takes over the `toEsQuery` functionality.

frontend/src/app/models/base-filter.ts
```diff
@@ -1,67 +1,76 @@
...
+
 /**
  * Abstract filter class
  *
  * Implements much of the logic to handle activity state.
+ *
+ * Filters are a StoreSync model, but have some wackiness because their internal state is
+ * not isomorphic to the stored state. In the UI, we distinguish between the default state
+ * (inactive, default data), and an inactive filter with non-default data. But this
+ * distinction is irrelevant for querying so it's not stored.
```
BeritJanssen, line 44: Words like "wacky" and "irrelevant" are not necessarily encouraging in docstrings. Can you rephrase?

frontend/src/app/utils/utils.ts
lukavdplas: I initially used it in several places and then removed all but one, so that's why it wasn't not defined inline.

jgonggrijp: I don't know whether Underscore or Lodash is used in I-analyzer, but this is `_.constant`. It is very general-purpose so I would say there is no shame in having it as a standalone function.

PR #1412, opened 2024-02-08
fix: move onLoading above canvas only

### LINKED ISSUE #1394: Loading Spinner of visualization components should block the canvas only
Right now, requesting a visualization from the backend will display the loading spinner, but this will block not only the canvas, but also input elements controlling normalization etc. This means the user cannot interact with controls anymore while the graph is loading, effectively making it impossible to cancel the running tasks.

It would be better if the loading spinner only blocked the canvas, so the user could keep interacting with the controls, which could cancel the running request for visualization data to the backend.

BeritJanssen:
Close #1394 .

### DIFF
--- a/frontend/src/app/visualization/barchart/barchart.directive.ts
+++ b/frontend/src/app/visualization/barchart/barchart.directive.ts
@@ -155,7 +155,6 @@ export abstract class BarchartDirective<
         chartDefault.plugins.tooltip.intersect = false;
     }
 
-    @HostBinding('class.is-loading')
     get isLoading() {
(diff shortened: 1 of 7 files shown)


lukavdplas (approved):

lukavdplas:
Code looks fine, but please verify that the tests succeed or re-open de PR in a branch with the `bugfix/*` prefix.

BeritJanssen:
Ah, so that's why the tests didn't trigger, of course...

BeritJanssen:
Close in favour of #1458.

PR #1444, opened 2024-02-09
Fix rendering of geopoint array

ar-jan:
The coordinates field was rendered as `[object Object]`.

The coordinates [should be](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/geo-point.html) `Geopoint expressed as an array with the format: [ lon, lat]`.

I'm leaving this in draft in case we want to use a different representation in the elasticsearch index, because the order of lat/lon is inconsistent in some of the current indexed data.

### DIFF
--- a/frontend/src/app/document-view/document-view.component.html
+++ b/frontend/src/app/document-view/document-view.component.html
@@ -23,9 +23,7 @@
                                 [attr.lang]="document.language(field)"
                                 [innerHtml]="formatInnerHtml(field)">
                             </td>
-                            <td *ngIf="isGeoPointField(field)">
-                                Lat: {{document.fieldValue(field).coordinates[1]}}, Lon: {{document.fieldValue(field).coordinates[0]}}
-                            </td>
+                            <td *ngIf="isGeoPointField(field)">{{displayGeoPointField(field)}}</td>
                             <td *ngIf="isUrlField(field)">
                                 <a href={{document.fieldValue(field)}} target="_blank">{{document.fieldValue(field)}}</a>
                             </td>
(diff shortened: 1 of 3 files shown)


BeritJanssen:
In principle, the coordinates are now saved as suggested in the documentation here:
```
PUT my-index-000001/_doc/1
{
  "text": "Geopoint as an object using GeoJSON format",
  "location": { 
    "type": "Point",
    "coordinates": [-71.34, 41.12]
  }
}
```
As discussed earlier, it would be best if the logic to request & convert data happened in the backend.

frontend/src/app/document-view/document-view.component.ts
```diff
@@ -71,6 +71,10 @@ export class DocumentViewComponent implements OnChanges {
         return field.name === 'url' || field.name.startsWith('url_');
     }
 
+    isGeoPointField(field: CorpusField) {
+        return field.name === 'coordinates';
```
BeritJanssen, line 75: I don't think it's wise to hard-code the name as `coordinates` here, but rather check if the field has the [geopoint field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html) as `.es_mapping`.

frontend/src/app/document-view/document-view.component.html
```diff
@@ -18,7 +18,10 @@
                     <ng-container *ngFor="let field of propertyFields">
                         <tr *ngIf="document.fieldValue(field)">
                             <th [iaBalloon]="field.description" iaBalloonPosition="up" iaBalloonLength="medium">{{field.displayName}}</th>
-                            <td *ngIf="!isUrlField(field)" data-test-field-value [innerHtml]="highlightedInnerHtml(field)"></td>
+                            <td *ngIf="!isUrlField(field) && !isGeoPointField(field)" data-test-field-value [innerHtml]="highlightedInnerHtml(field)"></td>
+                            <td *ngIf="isGeoPointField(field)">
+                                Lat: {{document.fieldValue(field).coordinates[1]}}, Lon: {{document.fieldValue(field).coordinates[0]}}
```
BeritJanssen, line 23: It would be more elegant to do this conversion in the typescript part, also to make sure we don't only process fields called `coordinates`.

ar-jan: Done, is DocumentViewComponent the place to do this?

BeritJanssen (approved):

PR #1448, opened 2024-02-13
versioning docs

lukavdplas:
Adds documentation on versioning. Notably, I've tried to describe when a new version should be a patch/minor/major release, which I've found to be unclear on past occasions. Like many of our applications, I-analyzer cannot strictly conform to semver because it doesn't have an API, so I think it's useful to add some clarification.

It's also intended to provide some clarity for anyone who wants to host I-analyzer themselves.

This update is prescriptive rather than descriptive; read it as "going forward, we should do it like this". It mostly describes how we're already doing things, but not always.

### DIFF
--- a/documentation/Versioning.md
+++ b/documentation/Versioning.md
@@ -1,0 +1,59 @@
+# Versioning
+
+This document describes how versions are labelled, when we bring out new versions of I-analyzer, and what you can expect when updating to a new version.
+
+If you're a developer, you can use this document to determine what the label should be for your new version. After that, walk through the checklist in [Making a release](./Making-a-release.md) to publish the new version.
+
+## Release labelling
+
+I-analyzer versions are based on [semantic versioning](https://semver.org/), so we distinguish between _major_, _minor_, and _patch_ releases. However, note that I-analyzer is primarily a user application and does not offer a public API, so it does _not_ conform to semantic versioning.
+
+Since I-analyzer is not centered on an API, it's not always obvious what is considered "breaking" change. The lists below gives a more detailed breakdown of what can be included in a patch, minor, or major release.
+
+A _patch_ release can include:
+
+- Fixes to bugs
+- Performance improvements
+- Changes to the backend API
+- Changes to the python or npm dependencies
+- Database migrations
+- Code quality improvements
+- Documentation improvements
+
+However, patch release cannot make meaningful changes to the frontend behaviour, deep routing in the frontend, or require updates in the server configuration.
+
+A _minor_ release can include:
+
+- Everything listed for patch releases
+- New functionality in the frontend
+- Layout changes to the frontend that don't remove functionality
+- Backwards compatible changes to routing in the frontend. A change is backwards compatible if URLs from older versions will still direct to the same content.
+- New corpus definitions
+- Changes to corpus definitions where updating existing definitions is not required, such as new options or source data types
+- Changes that require minor updates to the server configuration that are backwards compatible, such as a new Django setting or environment variable.
+
+A _major_ release can inclue:
+
+- Everything listed for minor releases
+- Removing functionality in the frontend
+- Backwards incompatible changes to routing in the frontend: URLs from older versions will no longer direct to the same content
+- Backwards incompatible changes to corpus definitions: these require updating existing definitions
+- Changes that require updating the server configuration in a way that is _not_ backwards compatible
+
+## When to make a release
+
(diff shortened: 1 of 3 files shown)


documentation/Versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+I-analyzer versions are based on [semantic versioning](https://semver.org/), so we distinguish between _major_, _minor_, and _patch_ releases. However, note that I-analyzer is primarily a user application and does not offer a public API, so it does _not_ conform to semantic versioning.
+
+Since I-analyzer is not centered on an API, it's not always obvious what is considered "breaking" change. The lists below gives a more detailed breakdown of what can be included in a patch, minor, or major release.
+
+A _patch_ release can include:
+
+- Fixes to bugs
+- Performance improvements
+- Changes to the backend API
```
jgonggrijp, line 17: This does not belong in a patch release in my opinion. It's either backwards-compatible and minor or backwards-incompatible and major.

jgonggrijp (commented):
I don't see why you could not adhere to semver if you don't have a public API. It just means that you have to define "breaking change" in a different way in order to decide when you're dealing with a major version upgrade. For applications, that most likely means that you break something for the user; either, they cannot do something anymore in the way they used to do it, or some piece of functionality disappears entirely. You already encoded that in the rules in the new documentation.

Besides, I think I-analyzer *does* have a public API. The public API of any web application consists of the set of all URLs that it responds to, both backend and frontend. So these URLs, and the functionality offered through them, should also be taken into account when you decide whether a change is breaking, an extension or just a fix.

lukavdplas:
> I don't see why you could not adhere to semver if you don't have a public API.

What I mean is that not declaring a public API directly contradicts https://semver.org/#spec-item-1 so you can't say "I-analyzer conforms to semver but it doesn't have a public API".

Besides, semver is, to my understanding, designed to streamline the management of software dependencies, so it's crucial for software libraries/services that other libraries may depend on. End-user applications don't always fit the same mould.

> For applications, that most likely means that you break something for the user

Keep in mind that if you feel the answer to a question is too obvious to document, it may be because you have more experience than others.

In any case, the purpose of this documentation is also to provide clarity beyond "most likely".

> Besides, I think I-analyzer does have a public API. The public API of any web application consists of the set of all URLs that it responds to, both backend and frontend. 

Declaring the backend API as a public API rather than an internal one would be a significant change in how the application is managed, that would go far beyond the scope of a documentation update. We're thinking of changing this, see https://github.com/UUDigitalHumanitieslab/I-analyzer/discussions/1430 if you have thoughts or suggestions.

On a design level, backend views are not really future-proof (they don't have to be, because we don't promise that), and actually implementing #1430 will probably mean making some breaking changes to the API _before_ we encourage others to depend on it.

In any case, I don't think it's out of place to create a web application that makes no guarantees to users that they can build applications on the backend API, for the same reason applications can make changes to HTML templates that break scrapers without considering those "breaking". Just because an interface can be accessed and parsed by software applications doesn't mean it has to be documented, presented and maintained as a public API.

In the case of I-analyzer, I think it's justifiable to declare the backend API as internal and say we don't provide support for the kind of use cases where a user would directly make HTTP requests to the backend. Supporting those use cases is a good idea, in my opinion, but it is not the current reality of the application.

jgonggrijp:
> > I don't see why you could not adhere to semver if you don't have a public API.
> 
> What I mean is that not declaring a public API directly contradicts https://semver.org/#spec-item-1 so you can't say "I-analyzer conforms to semver but it doesn't have a public API".

That item also contains the phrasing "or exist strictly in documentation". In other words, it does not have to be an actual API, it is whatever you describe as being the "interface" in a more abstract sense.

Which supports your point about the backend API, by the way, so point taken.

> > For applications, that most likely means that you break something for the user
> 
> Keep in mind that if you feel the answer to a question is too obvious to document, it may be because you have more experience than others.
> 
> In any case, the purpose of this documentation is also to provide clarity beyond "most likely".

Oh, I did not mean to suggest that the versioning scheme can be left undocumented. To the contrary, I think adding this documentation is a good idea! I am mostly happy with what you wrote, so I should perhaps have clicked on the "approve" button. I clicked "comment" because I did not feel entitled to a full-blown verdict.

lukavdplas:
Ah, thanks for clearing that up!

I don't think our frontend meets the requirement of "precise and comprehensive" documentation either, though, but I can see how you can apply the logic of semver in that scenario.

JeltevanBoheemen (approved):

PR #1450, opened 2024-02-14
Bugfix/parse weird doc ids

### LINKED ISSUE #1380: tag requests + document IDs containing slashes = carnage
### What went wrong?

I'm having some trouble with my development environment (since today).

Most notable:
- Frequent crashes: failing request, terminal flooded with errors, etc.
- Shutting down the frontend before the backend can cause the backend to freeze my machine. Didn't try it the other way around.

None of the above.

### Screenshot

screenshot of terminal

Console output:

screenshot of browser console

-  local server

### Version

develop at 7170f219453f9cee71112f5b3b9ad589ecba9f93

### Steps to reproduce

On my PC, i can reliably reproduce the issue as follows:

- start elasticsearch, start the backend with `yarn start-back`, and the frontend with `yarn start-front`
- open https://localhost:4200
- navigate to a corpus, then back the home page, then back to a corpus, then back the the home page again
- the page is extremely slow to load the list of corpora and often fails to load images
- after about 30 seconds, the backend terminal will start flooding errors (screenshot above)

But I'm not confident you'd get the same effect on another machine.

I updated python to 3.9 (from 3.8) this morning, which may be what causes the issue?

lukavdplas:
Fixes all immediate issues when document IDs contain "weird" characters such as slashes, which messed up url parsing (close #1380)

In the future, we could consider a requirement that document IDs should be slugs? But that would require re-indexing.

### DIFF
--- a/backend/tag/tests/test_views.py
+++ b/backend/tag/tests/test_views.py
@@ -95,8 +95,12 @@ def test_patch_document_tags(auth_client, auth_user_tag, mock_corpus, auth_user_
         { 'tags': [auth_user_tag.id] }
     )
 
+
     assert status.is_success(response.status_code)
     assert len(response.data['tags']) == auth_user_tag.count == 1
+
+    doc = TaggedDocument.objects.get(doc_id=new_doc)
+    assert doc.tags.count() == 1
 
     response = patch_request({
         'tags': []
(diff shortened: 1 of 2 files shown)


JeltevanBoheemen (approved):
It kinda hurts my eyes, but as we discussed, I see this as the best solution at the moment.

PR #1467, opened 2024-02-20
Refactor visualisation selection

### LINKED ISSUE #1371: Show a hint for ngram visualisation without query
Currently, if the user has not put in a query, the n-gram visualisation is completely hidden. I think it might be a good idea to not hide it, but instead prompt the user to put in a query first. What do y'all think?

### LINKED ISSUE #1374: Visualisations reset to default upon reloading / loading from URL
### What went wrong?

When navigating to any visualization via URL (i.e. also via reloading the page), it navigates to the default visualization, i.e. Number of Results compared by the default, usually Year or Date. 

I expected to get the visualisation I put in the URL.

### Screenshot

_No response_

- [X] https://peopleandparliament.hum.uu.nl
- [X] a local server

### Version

_No response_

### Steps to reproduce

Go to any visualisation that isn't the default and hit reload.

lukavdplas:
- part of #1403 : visualisation selection is now based on a `VisualizationSelector` class that uses the `StoreSync` functionality.
- close #1374 : reloading the visualisation page now works as it should, without lifecycle errors.
- if the user has not entered a search term, the term frequency and ngram visualisations are disabled rather than hidden (close #1371)
- dropdown items now support `[disabled]` input.
- error messages now have a header, which makes them a bit clearer.
- error messages don't replace the visualisation content, they just appear above it. This makes more sense to me, but I'm fine going back to the old behaviour.

### DIFF
--- a/frontend/src/app/models/visualization-selector.spec.ts
+++ b/frontend/src/app/models/visualization-selector.spec.ts
@@ -1,0 +1,78 @@
+import * as _ from 'lodash';
+import { mockCorpus3 } from '../../mock-data/corpus';
+import { SimpleStore } from '../store/simple-store';
+import { Store } from '../store/types';
+import { Corpus } from './corpus';
+import { QueryModel } from './query';
+import { VisualizationSelector } from './visualization-selector';
+
+describe('VisualizationSelector', () => {
+    let store: Store;
+    let corpus: Corpus;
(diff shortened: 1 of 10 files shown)


BeritJanssen (approved):
Love the cleanliness of the visualization component now!

Issue #1468, opened 2024-02-20
Use `Results` class for visualisations


lukavdplas:
[`Results`](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/frontend/src/app/models/results.ts) is an abstract class in the frontend that is designed to handle various types of results based on a query. A child class will have some parameters of its own (or none), and fetch results when the query or the parameters change.

This class is currently used to fetch the [(paged) document results for a query](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/frontend/src/app/models/page-results.ts), but it was written so that it could also be the basis for visualisations. The advantage of this is that `Results` includes some logic that won't need to be repeated for every visualisation. It also results in a more consistent flow in the application.

However, visualisations are complex modules that tend to include some complex logic of their own, so this refactor isn't trivial

To fully use the `Results` class, I would propose (in this order):

- [ ] a `FrequentWordsResults` class that generates results for the wordcloud.
- [ ] an `NgramResults` class that generates results for the ngram visualisation.
- [ ] a `DocumentFreqResult` class that generates a histogram of document frequency in relation to a keyword or date field. Create subclasses for those two scenarios. Used for the document frequency graph.
- [ ] a `TermFreqResult` class that generates the frequency of the search term in relation to a keyword or date field. Used for the term frequency graph.

Some considerations:
- The wordcloud is the easiest case, because it doesn't feature any additional parameters or extra logic.
- The ngram visualisation has a lot of parameters, but that should map well to how `Result` already works.
	- An additional complexity is the use of caching, but I don't think that should raise any particular difficulty. Write `NgramResult.fetch()` so it uses the cached data in the same way that `NGramComponent.loadGraph()` currently does.
	- Another issue is that you need to request that the visualisation task is cancelled when the model stops observing it - usually because the parameters change.
- The document frequency graphs have a straightforward fetching function, but for _date_ histograms, they have the complexity that we need to fetch higher-resolution data when zooming in (e.g. per month rather than per year), and project the data for the zoom window onto the current data in the graph. Not sure how to go about this. Create an additional `DocumentFreqResult` for the smaller interval?
- At this point, writing the `TermFreqResult` should be relatively straightforward. However, the barchart module is a beast - refactoring it to use the `DocumentFreqResult` or `TremFreqResult` is welcome but not trivial.

Bonus points:
[...]

PR #1475, opened 2024-02-23
use results model for wordcloud

lukavdplas:
Refactors the `WordCloudComponent` so it uses the `Results` class (part of #1468)


frontend/src/app/services/api.service.ts
```diff
@@ -139,9 +139,9 @@ export class ApiService {
     }
 
     // Visualization
-    public wordCloud(data: WordcloudParameters): Promise<AggregateResult[]> {
+    public wordCloud(data: WordcloudParameters): Observable<AggregateResult[]> {
```
JeltevanBoheemen, line 142: Let's make this a habit ;)

frontend/src/app/visualization/visualization.component.html
```diff
@@ -105,8 +105,8 @@
                         (error)="setErrorMessage($event)"
                         (totalTokenCountAvailable)="showTokenCountOption = $event"></ia-histogram>
                 </ng-container>
-                <ia-wordcloud *ngSwitchCase=" 'wordcloud' " [corpus]="corpus" [queryModel]="queryModel" [visualizedField]="visualizedField"
-                    [resultsCount]="resultsCount" [asTable]="freqtable" [palette]="palette"
+                <ia-wordcloud *ngSwitchCase=" 'wordcloud' " [queryModel]="queryModel"
```
JeltevanBoheemen, line 108: What's with the whitespace around `'wordcloud'`? Is that interpreted? Does it break if we remove it?

JeltevanBoheemen (approved):
This makes total sense. Small questions, but good job consider approved!

frontend/src/app/visualization/visualization.component.html
lukavdplas: Knowing I-analyzer, this is either strange formatting that was just never changed, or removing the space will somehow break all CSV downloads.

Probably the former.

PR #1476, opened 2024-02-28
document preview layout

### LINKED ISSUE #1383: Re-arrange document preview layout

The layout of the document previews is a bit "for historical reasons" in my opinion. 

image

The column on the right is designed to hold the large link to the scan, which makes sense for the times corpus but can result in a lot of empty space. The document previews are already quite narrow, so it would make more sense if the content could actually fill the whole width.

Another annoyance is that documents can be opened as either a popup or in a separate page, but you can't open the page from the preview: you always have to open the popup first. 

I'm thinking of a layout somewhat like this:

image

lukavdplas:
Changes the layout of document previews.

- content can use the full width of the preview
- preview includes a direct link to the document page
- no dead space for documents without an attached scan

close #1383

Screenshot:

screenshot of i-analyzer document preview

Document with scan image:

screenshot of i-analyzer document preview, showing "scan image" button

This also changes the relevance score to just show the percentage (rather than a meter), which I think is clearer.

screenshot of i-analyzer document page

### DIFF
--- a/frontend/src/app/document/document-preview/document-preview.component.ts
+++ b/frontend/src/app/document/document-preview/document-preview.component.ts
@@ -1,7 +1,7 @@
 import { Component, Input } from '@angular/core';
 import { FoundDocument } from '../../models';
 import { DocumentPage } from '../../models/document-page';
-import { documentIcons } from '../../shared/icons';
+import { actionIcons, documentIcons } from '../../shared/icons';
 
 @Component({
     selector: 'ia-document-preview',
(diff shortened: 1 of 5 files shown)


frontend/src/app/document/document-preview/document-preview.component.ts
```diff
@@ -12,8 +12,15 @@ export class DocumentPreviewComponent {
     @Input() document: FoundDocument;
     @Input() page: DocumentPage;
 
+    actionIcons = actionIcons;
     documentIcons = documentIcons;
 
+    get documentUrl(): string[] {
```
JeltevanBoheemen, line 18: Technically missing `void` return type, but I assume a situation with `this.document` doesn't really arise

JeltevanBoheemen (approved):
Much cleaner, I like it.

PR #1479, opened 2024-02-28
version in footer

### LINKED ISSUE #1160: Show version in footer
This is implemented in our cookiecutter app: https://github.com/UUDigitalHumanitieslab/cookiecutter-webapp-deluxe/pull/40

We can adapt this change to the I-analyzer frontend.

lukavdplas:
Adds the version number in the footer!

close #1160 

screenshot of i-analyzer footer showing the text "version 5.4.0 (source code)"

The `environment.prod.ts` for production environments still needs to be adjusted for the new version to show up; otherwise you'll see the same static link to this repository.

**Changes for deployment:** `environment.prod.ts` must be updated to include `sourceUrl` and `version`.


frontend/build/build-pre.js
```diff
@@ -0,0 +1,79 @@
...
+            }
+            if (stderr) {
+                reject(`stderr: ${stderr}`);
+                return;
+            }
+
+            // format is either:
+            // git@github.com:https://github.com/UUDigitalHumanitieslab/I-analyzer.git
+            // or https://github.com/UUDigitalHumanitieslab/I-analyzer.git
+            // or https://USERNAME:SECRET@github.com/UUDigitalHumanitieslab/I-analyzer.git/
```
BeritJanssen, line 41: Why support all of these?

lukavdplas:
Status: this works, except in docker. The issue happens because the frontend container will try to import the root `package.json` file (which contains the version number), but that container only has access to the frontend, so the import fails.

At the moment, it's not crucial that the version is retrieved in docker at all, because we don't use it in production, but the build should still complete.

frontend/build/build-pre.js
lukavdplas: Different git configurations, maybe?

lukavdplas:
You can outside of docker. This doesn't seem to work in the docker container because `package.json` is located in the repository root directory, so the pre-build script will look for a path like `../../package.json`, one level above `/frontend`. You could copy `package.json` to somewhere in the frontend when building the container, but it won't have the same path.

Alternatively, we could say that we should also write the version number in `/frontend/package.json`.

lukavdplas:
Okay, I "fixed" the docker issue. The footer will say `Version: undefined` in docker, but the build won't fail. As I said, we don't support Docker in production, so that's fine as far as I'm concerned.

Also simplified the pre-build script:
- The url to the remote repository is provided in the `environment.ts` file, rather than retrieved through git. Since this isn't the cookiecutter app, we already know the url of this repository, and if you fork I-analyzer, you'll almost certainly want to change the environment file anyway.
- The remote URL just links to the repository, rather than the specific hash of the commit. In the case of I-analyzer, we mostly add the link to Github so people can read about the project, make issues, or find our contact details, so linking to an old version can lead to misunderstandings.
- Removed the build time. I don't think it conveys useful information to an average user (who may not know what "build time" is).

frontend/package.json
```diff
@@ -4,11 +4,14 @@
     "ng": "ng",
     "start": "yarn serve",
     "start-docker": "ng serve --host 0.0.0.0 --port 4200 --proxy-config proxy.conf.docker.json",
+    "preserve": "yarn prebuild",
```
BeritJanssen, line 7: Why this wrapper around `yarn prebuild`? The command's name can be misleading, I first thought this would be for backing something up [cf. here](https://www.wordnik.com/words/preserve), so if you'd like to retain the pre-build / pre-serve distinction, perhaps introduce a hyphen?

BeritJanssen (approved):
I think this is a good solution. Indeed, no need to support version display in Docker. If we do at some point migrate to deployment with Docker, we can certainly find solutions for this (e.g., creating / copying information about the version on the frontend service prior to building).

BeritJanssen:
Should we still document the usage of the pre-build script?

frontend/package.json
lukavdplas: Yes, I was also confused by this. Though  the documentation on [pre & post scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts) does not suggest you could also use `pre-` as a prefix.

BeritJanssen: Ah, didn't realize this was an npm thing.

PR #1481, opened 2024-02-29
conditional download of nltk stopwords

### LINKED ISSUE #1377: Never-ending NLTK data check
### What went wrong?

Running `yarn start` from the repository root completely floods my terminal with 

```
[nltk_data] Downloading package stopwords to /home/luka/Documents/I-an
[nltk_data]     alyzer/I-analyzer/backend/addcorpus/nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
```

repeated ad infinitum.

I did not get the same issue when running `yarn start-back` and `yarn start-front` in separate terminals, though.

This isn't really a big deal but the command is kind of useless for development if it does not give sensible terminal output.

Strictly speaking, you would only need to check the nltk data once during startup, but it's possible that happens during runtime too. In case, when starting up the application, I would expect to see this message only once.

Perhaps the download (or verification that a download is not necessary) actually does happen only once and this is just some annoying interaction between the frontend and backend output.

### Screenshot

screenshot of terminal showing the quoted message being repeated many times

- a local server

### Version

develop at 7170f219453f9cee71112f5b3b9ad589ecba9f93

### Steps to reproduce

Start a development server by running `yarn start` from the repository root.

BeritJanssen:
This branch is meant to close #1377. @lukavdplas , can you check whether this fixes the issue for you? I've seen the nltk check pop up multiple times during testing, that seems to not be happening anymore now. I think it's basically caused by a call to `load_corpus_definition`, so I'd expect the check to be run for every corpus that has stopword removal or stemming assigned, not infinitely, so there might still be a leakage issue elsewhere. In that case, this is just a cosmetic fix.

### DIFF
--- a/backend/addcorpus/es_settings.py
+++ b/backend/addcorpus/es_settings.py
@@ -23,9 +23,12 @@ def get_language_key(language_code):
 
     return Language.make(language_code).display_name().lower()
 
+
 def get_nltk_stopwords(language_code):
-    nltk.download('stopwords', settings.NLTK_DATA_PATH)
     stopwords_dir = os.path.join(settings.NLTK_DATA_PATH, 'corpora', 'stopwords')
+    if not os.path.exists(stopwords_dir):
+        nltk.download('stopwords', settings.NLTK_DATA_PATH)
+
     languages = os.listdir(stopwords_dir)
     language = get_language_key(language_code)
 


lukavdplas (approved):
Yup, that seems to fix it!

Issue #1322, opened 2023-11-16
Citation documentation per corpus


lukavdplas:
**Is your feature request related to a problem? Please describe.**
Since I-analyzer aims to be accessible to researchers with any experience level, and presents a wide variety of media in the different corpora, users may be uncertain about how to cite documents. So the flow from seeing something in I-analyzer and bringing it up involves an extra step where you consult external sources.

I think I-analyzer has the potential to be a powerful help in making students and researchers aware of how to reference digital media, which is in line with our mission as digital humanists.

**Describe the solution you'd like**
Write a brief (markdown) documentation page per corpus that gives examples for, say, APA, MLA and bibtex.

**Additional context**
This is a dressed-down version of #726 : add documentation about citation to corpora, without generating citations for documents. This is trivial on a technical level, and the research effort would also be a necessary preparation of #726.

lukavdplas:
Adding citation documentation is now possible!

The DBNL and Troonredes corpora have citation documentation, but I want to add citation docs to a few more corpora before I close the issue. For example:

- [ ] Times
- [ ] Delpher
- [ ] Dutch parliamentary debates

Issue #1410, opened 2024-02-07
Schema for JSON/YAML corpus definition


lukavdplas:
This is a preparation for the following issues:

#981 : we want to make it possible to create a corpus by providing a JSON/YAML file as the definition (see that issue for a rough draft).

#982 : we want a frontend module to build or adjust a corpus definition using a form. It makes sense that the frontend will just create the JSON that can be parsed by #981 and send that over the API.

Some necessary prep work is to create a well-defined, well-documented schema for this file.

PR #1488, opened 2024-03-05
Feature/citation info per corpus

lukavdplas:
This adds the option to provide citation guidelines for a corpus (#1322).

Citation documentation is written as a markdown file. It uses the Django template language to render some values, namely the date of retrieval and the url of the site. Ideally, we'll eventually add the release number and date as well.

Added citation guides for the Troonredes and DBNL corpora.


backend/corpora/troonredes/citation/citation.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Other citation styles
+
+Some style guides recommend that a citation of a speech should include both the date and location. The location was not always the same, and is not listed on I-analyzer. You can look it up on [history of Prinsjesdag (website of the Dutch senate)](https://www.eerstekamer.nl/historische_gebeurtenis/geschiedenis_prinsjesdag).
+
+## Referencing word models
+
+While the text dataset of the troonredes is taken from [troonredes.nl](http://troonredes.nl), the word models were trained by the Centre for Digital Humanities at Utrecht University.
+
+If you are the word models interface of this corpus in your research, you can cite it by citing the webpage.
```
lukavdplas, line 45: This is mostly because we don't actually have a citable version of the model available anywhere. If the model was published through zenodo / yoda etc., we could also add that as a reference.

BeritJanssen: After re-training the word models on the current version, we should indeed publish them.

backend/addcorpus/models.py
```diff
@@ -73,6 +73,12 @@ class CorpusConfiguration(models.Model):
         validators=[validate_markdown_filename_extension],
         help_text='filename of the markdown documentation file for this corpus',
     )
+    citation_page = models.CharField(
+        max_length=128,
+        blank=True,
+        validators=[validate_markdown_filename_extension],
+        help_text='filename of the citation specification (in markdown) for this corpus',
+    )
```
BeritJanssen, line 81: Anticipating #1410 , would it not be more suitable to do this through a `FileField`? So that the citation info can be uploaded?

BeritJanssen (changes requested):
I like it! But I do wonder whether we want to add the citation files to version control, or if it wouldn't be more suitable to manage them through `FileField`s instead. The same goes for corpus descriptions, now that I think of it. We should discuss the best solution here - we already have many good `.md` files in version control, but would not want to have to do this step still on the developer side if an admin user can instead add this information through the Django admin interface. No need to solve this right in this branch though - so consider this "approved" if you don't think a `FileField` is a good solution at this stage.

backend/addcorpus/models.py
lukavdplas: I think it makes more sense to keep using the same system that we use for descriptions, images and word model documentation. If we want to adjust it later, we need to adjust those three fields anyway.

lukavdplas:
> I do wonder whether we want to add the citation files to version control

Hm, they're only in version control because the corpus definition is. If a user is uploading a JSON corpus definition to somewhere in the server's data directory (which is not version-controlled), you would store the citation page somewhere there as well.

Regarding (version-controlled) Python corpora: I do think it can make sense to keep some attributes outside of version control, but this mostly makes sense for properties like `data_directory` and `es_index` that are specific to the server. In the case of citation info, I'm not sure what the advantage would be. To avoid the templating logic?

BeritJanssen:
No, but I understood that corpus creation would happen through the admin interface, not through `ssh`ing into a server. So then it would make sense to add corpus definitions, documentation and citation info through the admin interface, through file upload.

lukavdplas:
Ahh, I see!

Agreed, we'll probably switch to using `FileField` at that time, but that would include a migration of the image, description, and model documentation fields, so migrating a fourth field won't really be any extra work if all those fields are using the same system.

Though keep in mind that `FileField` is just a pointer to a location on disk (with extra methods). By default, you'll use a location somewhere in `MEDIA_ROOT` which is not version-controlled, but it is possible to use a location in `/backend/corpora` instead.

Which is to say, you don't have to exclude a file from version control to store it in a `FileField`. (Though it's a good idea to treat the field as read-only in that case.)

BeritJanssen:
Ah, then we're on the same page. I was also thinking that we can point the upload directory to a path within this repository to combine content uploaded through git and through the admin interface. We'll probably still need to figure out a way then to push to git if content is added through the admin.

BeritJanssen (approved):

PR #1490, opened 2024-03-06
Add `word_model_path` to parliament corpora

BeritJanssen:
In the first instance, this PR is intended to activate the Irish word models, but it also prepares addition of word models for Canada, Denmark and Norway.


BeritJanssen:
NB: as I don't want to deploy changes from #1484 just yet, I will also make a hotfix branch of this, rebased on main.

lukavdplas (approved):

Issue #1117, opened 2023-05-17
Clarify fields named "url"


lukavdplas:
https://github.com/UUDigitalHumanitieslab/I-analyzer-django/pull/68#discussion_r1196367632

The addition of unique links for documents (#1083) can cause some confusion when a document also has a field named "URL", which is true for several corpora. Such fields usually refer to the source data.

We can update the display names for extra clarity, e.g. include the name of the database or the type of file this is linking to.

PR #1491, opened 2024-03-06
clarify fields named "URL"

lukavdplas:
See #1117  : naming a field "URL" in the interface is unclear because I-analyzer also creates a unique URL for the document. This update clarifies the display names for most of those fields, by making it clear it's the URL _on Goodreads_ (or wherever).

The exception is this one:

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/7b84d6b93d1876e5834e933249ca8c04ab7d89ba/backend/corpora/peaceportal/utils/field_defaults.py#L29-L36

where I don't know what it's supposed to link to. @BeritJanssen , could you add this?

### DIFF
--- a/backend/corpora/peaceportal/utils/field_defaults.py
+++ b/backend/corpora/peaceportal/utils/field_defaults.py
@@ -29,7 +29,7 @@ def source_database():
 def url():
     return FieldDefinition(
         name='url',
-        display_name='URL',
+        display_name='Source URL',
         description='URL of the inscription entry.',
         es_mapping=keyword_mapping(),
         search_field_core=True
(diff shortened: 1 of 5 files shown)


JeltevanBoheemen:
How about `source`? as a field name?

lukavdplas:
Oh, good one! I decided to go with `Source URL` as the display name, I think that's clear.

JeltevanBoheemen (approved):

BeritJanssen:
Should I push to this branch to add the Peaceportal info?

backend/corpora/peaceportal/utils/field_defaults.py
```diff
@@ -29,7 +29,7 @@ def source_database():
 def url():
     return FieldDefinition(
         name='url',
-        display_name='URL',
+        display_name='Source URL',
         description='URL of the inscription entry.',
```
BeritJanssen, line 33:
```suggestion
        description='URL of the inscription entry in the source database.',
```

BeritJanssen: Perhaps this clarifies it somewhat?

PR #1492, opened 2024-03-06
fix: trigger prebuild before starting frontend server

BeritJanssen:
This adjustment of `docker-compose` is still necessary in order to compile in Docker after #1479 .


docker-compose.yaml
```diff
@@ -50,7 +50,7 @@ services:
       - type: bind
         source: ./frontend/build
         target: /frontend/build
-    command: sh -c "yarn start-docker"
+    command: sh -c "yarn prebuild && yarn start-docker"
```
lukavdplas, line 53: Doesn't `yarn start-docker` redirect to `ng serve`/`ng build`? If so, `yarn preserve`/`yarn prebuild` should run implicitly.

BeritJanssen: It does not do `yarn serve` but `ng serve`, so that's why (I think) the preserve is not triggered.

PR #1497, opened 2024-03-08
Feature/json schema for corpora

### LINKED ISSUE #1410: Schema for JSON/YAML corpus definition

lukavdplas:
This adds a JSON schema for corpus definitions (close #1410). The schema can't be imported yet.

- Adds a JSON schema for corpus definitions.
- Adds a `validate()` function that checks whether a JSON corpus definition conforms to the schema
- Adds a JSON definition of the Goodreads corpus that conforms to the schema.
- Adss a unit test that validates `goodreads.json`.


backend/addcorpus/schemas/corpus.schema.json
```diff
@@ -0,0 +1,226 @@
...
+                        "delimiter": {
+                            "type": "string",
+                            "description": "delimiter between values in the source files",
+                            "enum": [",", ";", "\t"]
+                        }
+                    }
+                }
+            }
+        },
+        "fields": {
```
lukavdplas, line 92: I merged these files to we just have a single `corpus.schema.json`, rather than using the URI of `field.schema.json`, because it's easier to implement than referencing a URI. Also, the resulting file isn't really that big.

JeltevanBoheemen (approved):
Nice!

PR #1498, opened 2024-03-08
Feature/unauthenticated download

### LINKED ISSUE #1229: Enable downloads without user accounts
The "lowest hanging fruit" solution for non-authenticated users is to enable the "direct" download option. That would mean we'd have a rather low max download size (1000). Will file another issue for more flexible solutions.

BeritJanssen:
Close #1229. Non-authenticated users can make use of the "direct" download functionality.


frontend/src/app/search/search.component.html
```diff
@@ -36,7 +36,7 @@
                                 </div>
                             </form>
                         </div>
-                        <ia-download [corpus]="corpus" [queryModel]="queryModel" [resultOverview]="resultOverview" [hasLimitedResults]="hasLimitedResults" [downloadLimit]="user.downloadLimit" [route]="this.router.url" *ngIf="user"></ia-download>
+                        <ia-download [corpus]="corpus" [queryModel]="queryModel" [resultOverview]="resultOverview" [hasLimitedResults]="hasLimitedResults" [downloadLimit]="user? user.downloadLimit: 1000" [route]="this.router.url"></ia-download>
```
JeltevanBoheemen, line 39: I'd like to move the fallback value of `downloadLimit` to some centrally located constant if possible.

JeltevanBoheemen (commented):
Nice! One suggestion.

BeritJanssen:
@lukavdplas, I introduced a `direct_download_limit` on the corpus. Can you have a quick look if you agree with this solution? Advantage: no hard-coded values anymore. Disadvantage: one extra setting, which may get confusing if we also want to impose a `download_limit` for a corpus.

BeritJanssen:
The reasoning to do it per corpus was that it was easy to send it along with the corpus request, instead of still having to fire off a dedicated request & create an extra view (because so far we don't have such a "general information" call, as far as I'm aware). I reasoned that it would also be easy this way to override for the PeacePortal corpora. But then, a server-wide setting would definitely also work for that.

lukavdplas:
I think that if you don't want to have different settings per corpus on the same server, it shouldn't be in the database.

As for the "general information" call: it's not impossible to include some field from the settings in the corpus endpoint (instead of retrieving the value from the model), but we usually handle those kind of settings by also including them in `environment.ts`.

BeritJanssen:
Yes, also thinking that direction now: should add the direct download limit to the backends `settings.py` and the frontend's `environment.ts`. Duplicating information in this repo, but for deployment, it can be set from one value. And less involved than sending this information along with existing requests, let alone create a new one.

PR #1500, opened 2024-03-08
enable caching of GitHub actions

### LINKED ISSUE #1489: Github actions don't cache
We should look into a way to cache the Docker images built for the actions. Right now, they are re-built every time, which is rather inefficient.

BeritJanssen:
Close #1489 .


BeritJanssen:
I'll start this over on another branch.

Issue #1333, opened 2023-11-24
Corpus validation after all fields are added


lukavdplas:
**Is your feature request related to a problem? Please describe.**
We recently added a LOT of validation for corpora, which is nice. But some requirements for fields can't be validated because you need to be sure all fields have been added. For example, this docstring suggests that there must be an 'id' field ,but that is never enforced: 

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/9224528af4a435b84d8708fdad422eb8f2980ac9/backend/addcorpus/corpus.py#L109 

There are some similar requirements, like how each corpus should at least have one `text_content` field.

We *can't* validate this at the moment, because you need to save the `CorpusConfiguration` in the databse before you can add fields to it. You also can't add this to field validation, because fields are saved one by one.

This also applies to #982 - a user should be able to save their "working version" of a corpus before they have added every single field.

**Describe the solution you'd like**
Perhaps we can add a second validation method that isn't run when saving the model, but *is* run when you want to index the corpus or include it in the interface. So you can save a corpus without an ID field, but I-analyzer will consider it a draft and refuse to include it in the interface.

PR #1501, opened 2024-03-11
Feature/default sort

lukavdplas:
Changes the way that the default sort setting is represented. This is now a `default_sort` attribute of the _corpus_, rather than an attribute of each field.

This is a small preparation for importing the JSON definition in #1497 . It's not strictly necessary to import the JSON, but the frontend _also_ stores the default sort configuration on the corpus rather than the field, so this update means there will be less conversion overall.

The field is provided as a string (in execution order, `default_sort` may be defined before `fields`). There is no validation whether the field exists and is sortable - that depends on #1333


JeltevanBoheemen (approved):
This makes total sense. Clean code.

PR #1502, opened 2024-03-11
Feature/corpus module structure

lukavdplas:
This moves a few of the files in `backend/addcorpus` to a subfolder named `python_corpora`.

These files contain base classes that are used in Python corpus definitions, or functionality for importing them. Since these will no longer be the only method of adding corpora, it makes sense that this exists in its own module.

Note for review: it's only necessary to check the changes in `backend/addcorpus`. Everything outside is just import updates.


JeltevanBoheemen (approved):
ALl I'm seeing is import changes, in `backend/addcorpus` too. Good import changes btw!

Issue #981, opened 2022-11-25
Import/export corpus definitions


lukavdplas:
There should be a method to import/export corpus definitions, where the file and database have a one-to-one mapping. Export could be json or yaml (or something else?).

This is useful for offering user support, debugging, transferring data between local, test, acc and production servers. It can also be useful to curators who want to create different version of a corpus.

This should be accessible from the interface (for curators), but would also be useful as a command line tool.

For development purposes, we should probably develop this this _before_ the option to enter corpus definitions via a form, as the latter will take up much more development time.

We can consider if we want a centralised location to save exported corpus definitions. This could be the current `corpora` module, but perhaps it makes more sense to make this a separate repository.

To do:
- [ ] Function to import a JSON definition (#1519)
- [ ] Function to export a JSON definitio (#1547)

lukavdplas:
What is the easiest way to make the JSON import/export usable in practice?

In the short term, we want to be able to use this import/export to test things out. In the long term, we do need an easy way to import/export the JSON-serialised corpus, since that makes it easy to transfer corpora between servers.

Options:

- Add this to the admin site. This is a good place to have such an action, but I'm not very familiar with customising the Django admin like this. Perhaps as an [action that provides an intermediate page](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages)? That looks appropriate for an export option (that's actually the example in the documentation), but I'm not sure how an import could work.
- Use [Django REST framework's API interface](https://ianalyzer.hum.uu.nl/api/). Since the frontend will use the same JSON schema, this can be recycled as the actual view.
- Add a [custom command](https://docs.djangoproject.com/en/5.0/howto/custom-management-commands/)
- Build a (very rudimentary) frontend component with just the JSON import/export. Can be expanded with a proper form later.
- ?

PR #1504, opened 2024-03-12
corpus validation at different stages

### LINKED ISSUE #1333: Corpus validation after all fields are added

lukavdplas:
This update changes the validation for corpora. The idea is that corpora have three "checks" available for different stages in their lifecycle:

- Validate that a corpus can be saved in the database.
- Validate that a corpus can be indexed.
- Validate that a corpus can be used in the interface.

The first is the normal `clean()` method (and is mostly unchanged). Corpora now have two additional checks, `ready_to_index()` and `ready_to_publish()`.

Also adds an `active` field to corpora. Right now, it just functions as a stored value of `ready_to_publish()` (for efficiency).

Close #1333

It doesn't add much validation yet. Corpora are now required to:

- have at least one content and one metadata field in order to index
- have a sensible value for `default_sort` in order to be public

I expect we will expand the validation as part of #981. The purpose of this update is mainly so we can start to sort those checks into the correct stage.


backend/addcorpus/models.py
```diff
@@ -46,6 +60,66 @@ class Meta:
...
+        '''
+        Checks whether the corpus is ready for indexing.
+
+        Runs a try/except around `self.validate_ready_to_index()` and returns a
+        boolean; `True` means the validation completed without errors.
+
+        If you want to see validation error messages, use the validation method
+        directly.
+        '''
+        return _validation_passes(self.validate_ready_to_index)
```
JeltevanBoheemen, line 74: Hmm this seems convoluted. Why not just make `_validation_passes` raise a `ValidatonError`?

backend/addcorpus/models.py
```diff
@@ -46,6 +60,66 @@ class Meta:
...
+
+        validate_essential_fields(fields)
+
+
+    @admin.display()
+    def ready_to_publish(self) -> bool:
+        '''
+        Checks whether the corpus is ready to be made public.
+        '''
+        return _validation_passes(self.validate_ready_to_publish)
```
JeltevanBoheemen, line 98: Tying in to the comment above, I would place the `try/except` here, and return `True/False`

backend/addcorpus/python_corpora/save_corpus.py
```diff
@@ -87,17 +87,26 @@ def _save_field_in_database(field_definition: FieldDefinition, configuration: Co
     field.full_clean()
     return field
 
-def _set_corpus_inactive(corpus_name):
+def _deactivate(corpus):
+    corpus.active = False
+    corpus.save()
+
+def _set_active(corpus):
```
JeltevanBoheemen, line 94: These method names don't match. 
```suggestion
def _deactivate(corpus):
    corpus.active = False
    corpus.save()

def _activate(corpus):
```

backend/addcorpus/validation/publishing.py
```diff
@@ -0,0 +1,45 @@
+'''
+This module defines functions to check if a corpus is ready to be published.
```
JeltevanBoheemen, line 2: Sweet!

backend/addcorpus/validation/indexing.py
```diff
@@ -0,0 +1,53 @@
+'''
+This module defines functions to check if a corpus is ready for indexing.
```
JeltevanBoheemen, line 2: Sweet!

JeltevanBoheemen (commented):
As always, well written, well documenten. Great job!
Minor remarks.

backend/addcorpus/models.py
lukavdplas: I agree outfactoring `_validation_passes` was a bit unnecessary, if that's what you mean.

The reason why it this method doesn't raise `ValidationError` is that I thought it would be useful to have a method that does not raise errors but just says whether validation passes.

Basically, there are some situations (like trying to index) where you really should raise errors if the corpus is not index-ready, but if you're just displaying the state of the corpus (e.g. in the admin), it's not an "error" if the corpus is incomplete.

PR #1514, opened 2024-03-18
no load _corpus_definition in wordcloud

lukavdplas:
An update to improve compatibility with database-only corpora:  this removes the call to `load_corpus_definition` in the runtime of the wordcloud.

Stopwords are now based on the `language` property of the field (as saved in the database), rather than the configured analyser in elasticsearch.

Also made sure all fields in existing corproa with language-specific analysers actually set the `language` property.


backend/corpora/parliament/finland-old.py
```diff
@@ -67,7 +67,7 @@ def sources(self, start, end):
     source_archive = field_defaults.source_archive()
     source_archive.extractor = CSV(field='file')
 
-    speech = field_defaults.speech()
+    speech = field_defaults.speech(language='fin')
```
BeritJanssen, line 70: I believe 'fi' is the language string? Though 'fin' may also be valid. Have you tested with the `langcodes` module?

BeritJanssen (changes requested):
Looks good, apart from a comment wrt `finland` / `finland-old`. As for the bilingual nature of corpora, `canada` would also seem a candidate to perhaps do without a language key, though I seem to remember that the portion of French in that corpus was negligible. NB, I believe that with the current field default being `stopword_analysis=True`, we'd run into an error with an empty language field? So perhaps some work in `addcorpus.es_mappings` is required. Perhaps this can also split out into a separate issue. As reindexing of parliament corpora is not going to happen in the very near future, this problem doesn't need to be addressed in this PR.

backend/visualization/wordcloud.py
```diff
@@ -1,19 +1,17 @@
...
-    field_definition = next((f for f in corpus.fields if f.name == field), None)
-    mapping = field_definition.es_mapping
-    analyzer = mapping.get(
-        'fields', {}).get('clean', {}).get('analyzer')
-    if not analyzer:
+def field_stopwords(corpus_name, field_name):
+    corpus = Corpus.objects.get(name=corpus_name)
+    field = corpus.configuration.fields.get(name=field_name)
+    if field.language and field.language is not 'dynamic':
+        return get_nltk_stopwords(field.language)
```
BeritJanssen, line 12: Ah, I see now why you'd want the `field.language` to be set. For mixed corpora, this would fall back to the first language of the `corpus.languages` array? So we'd still remove Finnish stopwords from Swedish texts. I do think we should also have an option to skip stopword removal altogether, but this can be left for later.

backend/corpora/parliament/finland-old.py
lukavdplas: It's `fi` in ISO 631-1 and `fin` in 639-3. I think `langcodes` accepts either, but we don't apply standardisation to the names of analysers in elasticsearch, so this could cause issues.

In any case, I didn't know there was also a lot of Swedish, so you're right, the language should not be set like this.

backend/visualization/wordcloud.py
lukavdplas: In the wordcloud here, there is no fallback, so if the field has no language set (or it's "dynamic", meaning it's mixed but the value is included in the document), you don't remove stopwords.

I think that's more appropriate as the default behaviour, but I can imagine that for some corpora, it would be acceptable to remove stopwords based on the majority language.

lukavdplas: If you're really ambitious, the `'dynamic'` setting could also be used to remove finnish stopwords in finnish documents, and swedish stopwords in swedish documents.

That's a lot more complicated though, so it's not really in the scope of this PR.

BeritJanssen: Hm, I'm getting a bit confused by `field.language` (see my [question](https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1522)), because the code here seems to indicate it can be an empty string, while the validator associated with the field seems to throw a ValidationError for values which are not "dynamic", or a valid language code.

backend/corpora/parliament/finland-old.py
BeritJanssen: You would need to override the field default of `es_mappings` though, since that has stopword removal set to `True`, which would fail without the language information.

backend/visualization/wordcloud.py
BeritJanssen: I totally agree with *not* getting overcomplicated here with stopword removal, and just leave that step our for mixed corpora.

PR #1515, opened 2024-03-18
implement has_python_definition on corpus model

lukavdplas:
The intention of this update is to make the corpus database a bit more robust when we're adding database-only corpora. 

This adds a boolean property `has_python_definition` for corpora. It's not used anywhere, but it could be used to:

- block changing the corpus through the admin
- disable "advanced" features
- check whether a property should be auto-generated or read from the definition class.

This also adjusts the logic for the `save_corpus` module (saving python corpora into the database). 

First, the `has_python_definition` property is used as follows:

- If a corpus definition is successfully loaded from settings, it will set `active = True` and `has_python_definition = True`.
- If a corpus definition is included in settings but can't be imported, the corpus will set `active = False` and `has_python_definition = True`.
- If a corpus is _not_ in the settings and has `has_python_definition=True`, it will set `has_python_definition = False` and `active = False`. At that point, you could remove it or use it as a database-only corpus.
- If a corpus is _not_ in the settings and has `has_python_definition=False`, nothing happens.

Another difference is that `CorpusConfiguration` objects are a bit more stable:

- If a corpus is removed from settings or can't be loaded, the configuration is not deleted. (But the corpus will be inactive.)
- When updating a corpus, the existing configuration is updated, rather than deleted and re-created.


backend/addcorpus/admin.py
```diff
@@ -81,7 +82,8 @@ class CorpusConfigurationAdmin(admin.ModelAdmin):
     ]
 
     def get_form(self, request, obj=None, **kwargs):
-        show_warning_message(request)
+        if obj and obj.corpus.has_python_definition:
+            show_warning_message(request)
```
JeltevanBoheemen, line 86: Neat

backend/addcorpus/migrations/0014_corpus_has_python_definition.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('addcorpus', '0013_alter_field_visualizations_map'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='corpus',
+            name='has_python_definition',
+            field=models.BooleanField(default=False, help_text='whether the configuration of this corpus is determined by a Python module (some features are only available for Python-based corpora)'),
```
JeltevanBoheemen, line 16: Should this contain a data migration?
What happens when we (for some reason) need to roll back a few migrations, and then reapply this. Existing corpora that have set `has_python_definition=True` will then be `False` because of the default.

backend/addcorpus/python_corpora/save_corpus.py
```diff
@@ -4,16 +4,27 @@
 from addcorpus.python_corpora.load_corpus import load_all_corpus_definitions
 import sys
 
+def _configuration_pk(corpus: Corpus):
```
JeltevanBoheemen, line 7: I would put this as a getter in the `Corpus` class

backend/addcorpus/python_corpora/save_corpus.py
```diff
@@ -63,6 +74,14 @@ def _save_corpus_fields_in_database(corpus_definition: CorpusDefinition, configu
     for field in corpus_definition.fields:
         _save_field_in_database(field, configuration)
 
+    for field in configuration.fields.exclude(name__in=corpus_definition.fieldnames):
+        field.delete()
+
+def _field_pk(name: str, configuration: CorpusConfiguration):
+    if Field.objects.filter(corpus_configuration=configuration, name=name).exists():
+        field = Field.objects.get(corpus_configuration=configuration, name=name)
```
JeltevanBoheemen, line 82: Double lookup, prefer `try/except Field.DoesNotExist`

JeltevanBoheemen:
```suggestion
    try:
        return Field.objects.get(corpus_configuration=configuration, name=name).pk
    except Field.DoesNotExist:
        return None
```

backend/addcorpus/python_corpora/save_corpus.py
```diff
@@ -101,14 +122,14 @@ def _activate_if_ready(corpus):
     corpus.active = corpus.ready_to_publish()
     corpus.save()
 
-def _clear_configuration(corpus):
+def _clear_python_definition(corpus):
     '''
-    Remove the configuration attached to a corpus.
+    Mark a corpus as one without a python definition deactivate it.
```
JeltevanBoheemen, line 127:
```suggestion
    Mark a corpus as one without a python definition and deactivate it.
```

JeltevanBoheemen (approved):
The goal of the PR is clear and achieved.
The logic surrounding corpus definitions is getting complex, but I think you managed to keep it clear. 
Few non-blocking code remarks.

backend/addcorpus/migrations/0014_corpus_has_python_definition.py
lukavdplas: Yes, I wasn't sure about this step.

I think in that situation, the value would be set to `False` in the migration, but set to `True` when you import Python corpora (which happens when you start the server).

JeltevanBoheemen: Oh yeah, that makes sense. In that case it's acceptable to leave the data migration out. Maybe provide a comment that this is done implicitly at server start.

PR #1519, opened 2024-03-20
parse JSON corpus definition

### LINKED ISSUE #978: Save complete corpus definition in database
In order for curators to add corpora via the interface, it must be possible to save a complete definition of a corpus in the SQL database. Currently, the Corpus table underspecifies the corpus, and most information is stored in the python files.

lukavdplas:
Adds a function to import a JSON corpus definition into the database (resolves #978 and half of #981)

The import function saves a `Corpus` object in the database, though the function is never called (outside of tests).

Some functionality is still not supported for the saved corpus, most notably:
- Extracting source data / indexing
- Imported range filters and date filters won't work since they miss an upper/lower bound (#1535). For now, you need to add those values in the admin interface if you want to use the corpus.

Some extra additions to make this work:
- Added some extra logic to handle unsupported languages when configuring a main content field. Analysers are only added when they're available. This is also applied when Python corpora use the utility functions for elasticsearch mappings + settings.
- Added an (optional) `index_prefix` option to the configuration of an elasticsearh server. So a new corpus can get an index name like `ianalyzer-mycorpus`.
- Added extra fields to the database for extraction data.


backend/addcorpus/json_corpora/import_json.py
```diff
@@ -0,0 +1,282 @@
+from typing import Dict, Iterable
+from datetime import datetime
+
+
+from addcorpus.models import Corpus, CorpusConfiguration, Field
+from addcorpus.json_corpora.utils import get_path
```
BeritJanssen, line 6: where is the utils module? Did you forget to commit it?

backend/addcorpus/json_corpora/import_json.py
```diff
@@ -0,0 +1,282 @@
...
+    field.language = language
+    field.display_type = 'text_content'
+    field.search_filter = {}
+    field.searchable = True
+    field.search_field_core = True
+
+    visualize = get_path(field_data, 'options', 'visualize')
+    if visualize:
+        field.visualizations = [
+            VisualizationType.WORDCLOUD.value
```
BeritJanssen, line 131: Why is this set to WORDCLOUD now?

backend/addcorpus/json_corpora/import_json.py
lukavdplas: It's in the files changed under `/backend/addcorpus/json_corpora`? I don't know if I understand the issue.

backend/addcorpus/json_corpora/import_json.py
lukavdplas: Yes, it's a boolean: https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/1dfa42b8aafab54108fa8e03bb43a7ace6527be7/backend/addcorpus/schemas/corpus.schema.json#L140-L143

backend/addcorpus/json_corpora/import_json.py
```diff
@@ -0,0 +1,282 @@
...
+    field.sortable = get_path(field_data, 'options', 'sort')
+    filter_setting = get_path(field_data, 'options', 'filter')
+
+    if filter_setting == 'show':
+        field.search_filter = {
+            'name': 'DateFilter',
+            'description': f'Select results based on {field.display_name}',
+        }
+    else:
+        field.search_filter = {}
```
BeritJanssen, line 230: Why isn't it possible to read out the min/max date values of the corpus here? Are those optional?

lukavdplas: They're not specified in the JSON (since it's a simpler version).

lukavdplas:
@JeltevanBoheemen this branch is out of date with develop. If we bring it up to date, it will cause a migration conflict (addcorpus has one new migration on develop from #1531, and this branch adds migrations too). I know you already applied the migrations on this branch locally, can you update the branch and "rebase" the migrations?

jgonggrijp:
Did you know that `django makemigrations` has a `--merge` option?

https://docs.djangoproject.com/en/4.2/ref/django-admin/#cmdoption-makemigrations-merge

lukavdplas:
Yes, I've been asked not to use those in feature branches and redo the migrations instead.

jgonggrijp:
Why? By whom?

lukavdplas:
By Berit, because the migration history is easier to read if you don't have merge migrations.

backend/addcorpus/migrations/0005_add_validators.py
```diff
@@ -4,6 +4,11 @@
 from django.db import migrations, models
 
 
+def validate_image_filename_extension(filename):
+    allowed = ['.jpeg', '.jpg', '.png', '.JPG']
```
BeritJanssen, line 8: what about other filename extensions, for instance `webp`?

backend/addcorpus/models.py
```diff
@@ -227,6 +229,12 @@ class CorpusConfiguration(models.Model):
         help_text='name of the field that specifies the language of documents (if any);'
             'required to use "dynamic" language on fields',
     )
+    source_data = models.JSONField(
```
BeritJanssen, line 232: Why use a `JSONField` here and for `extract_options`. Would it not be desirable to limit and validate choices? E.g., that this could be `csv/json/xml/` etc.? From the name, I actually also expected `source_data` to define the location of the source data, not its format, so perhaps renaming to `source_data_format` would be clearer.

backend/addcorpus/validation/publishing.py
```diff
@@ -2,6 +2,7 @@
 This module defines functions to check if a corpus is ready to be published.
 '''
 
+import os
```
BeritJanssen, line 5: unused import?

backend/ianalyzer/common_settings.py
```diff
@@ -130,3 +130,5 @@
 LOGO_LINK = 'https://dhstatic.hum.uu.nl/logo-cdh/png/UU_CDH_logo_EN_whiteFC.png'
 
 NLTK_DATA_PATH = os.path.join(BASE_DIR, 'addcorpus', 'nltk_data')
+
+DEFAULT_CORPUS_IMAGE = os.path.join(BASE_DIR, 'addcorpus', 'images', 'default.png')
```
BeritJanssen, line 134: Where it this setting used?

backend/addcorpus/validation/creation.py
```diff
@@ -186,3 +188,25 @@ def validate_sort_configuration(sort_config):
 
     if type(ascending) is not bool:
         raise ValidationError(f'Sort configuration has invalid "ascending" property: {ascending}')
+
+
+def validate_source_data_configuration(source_data: Dict) -> None:
```
BeritJanssen, line 193: Please provide a docstring. What are expected inputs? In which cases should the ValidationError be raised?

backend/addcorpus/validation/creation.py
```diff
@@ -186,3 +188,25 @@ def validate_sort_configuration(sort_config):
...
+    if source_data == {}:
+        return
+    try:
+        validate(source_data, 'properties', 'source_data')
+    except JSONSchemaValidationError as e:
+        raise ValidationError(
+            ['Corpus source data schema validation failed', e])
+
+
+def validate_field_extract_options(extract_options: Dict) -> None:
```
BeritJanssen, line 203: Same as above.

BeritJanssen (changes requested):
Nice! Some minor things:
- the purpose of the `source_data` and `extract_options` fields and associated validators need more documentation
- there is a default corpus image now, but I don't see it used anywhere. It's very likely I overlooked this, but please check and/or document what the default corpus image should be used for.

backend/addcorpus/migrations/0005_add_validators.py
lukavdplas: Yes, this function is quite rudimentary, but it isn't used anymore, which is why it was moved to an older migration file. (See https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/1526#discussion_r1557321862)

backend/addcorpus/models.py
lukavdplas: I think I agree. If we add this data as flat fields, we only need `source_data_delimiter` on the CorpusConfiguration and `extract_options_column` on the Field model. Both would be `CharField`. (File type is included in the JSON for forwards compatibilty, but there is only one option.)

@JeltevanBoheemen what do you think?

JeltevanBoheemen: We chose a `JSONField` to be flexible for future extensions to both set of options, which is a very conceivable situation. 
The described validation is applied, through the JSON schema. If we support more source_data_formats, it would simply be a matter of adding them to the enum in the schema. 
I still stand behind this implementation, but agree that we should document it better.

lukavdplas: I agree that the JSON field already has fine validation, but what convinced me is that JSON fields are difficult to work with in the Django admin. If you encode the actual options, the delimiter can just be a dropdown, instead of nested in a JSON. I do think having a nice(-ish) admin form is a good intermediary goal before we build the frontend.

Also, if we do add more options, we can always migrate the existing data into a JSON field.

JeltevanBoheemen: Point taken on the admin interface

PR #1521, opened 2024-03-21
Feature/hide tags

BeritJanssen:
Close #1484 .

### DIFF
--- a/frontend/src/app/document-view/document-view.component.html
+++ b/frontend/src/app/document-view/document-view.component.html
@@ -9,7 +9,7 @@
                             <ia-search-relevance [value]="document.relevance"></ia-search-relevance>
                         </td>
                     </tr>
-                    <tr>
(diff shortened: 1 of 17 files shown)


JeltevanBoheemen (changes requested):
I feel a little uncomfortable writing a review like this, but I don't agree with the general direction of this implementation. I think including authentication information in the `QueryModel` is unnecessary and overcomplicated. Instead, I propose the following general idea:

- `TagService` keeps track of authentication state, setting a property if user is not authenticated (e.g. `taggingAvailable`)
- If this is difficult to set usng the `authService`, the view that returns tags can raise a `HTTP401` and the `TagService` can respond to this specific error response, setting aforementioned property.
- Interface elements that use tags in some way check for this property. This way, it is rather easy to make the distinction between not having tags (aka empty list) or not having tags enabled. 

Of course, feel free to discuss!

Issue #1524, opened 2024-03-22
Change data model for corpus attachments


lukavdplas:
Currently, corpus definitions used "attached files", namely the image and various documentation files. The model is that the corpus definition specifies the filename, and this is resolved by looking in the directory that contains the corpus definition.

This system won't work for database-only corpora, so it should be adjusted.

### Corpus images

Use an `ImageField` to save the image file. The ImageField is a specialised version of `FileField`. See the [Django documentation on file management](https://docs.djangoproject.com/en/5.0/topics/files/).

Note: it's theoretically possible to have the ImageField point to the original file in the source code, but this creates a mess when some corpora don't save their data in the source code. I think it's better to save all data in MEDIA_ROOT.

### Documentation files

Create a model `CorpusDocument` with the following fields:

- `corpus_configuration`: a `ForeignKey` to the CorpusConfiguration.
- `type`: a field with `choices` for the type of document: general information, word models documentation, citation info. Possibly new categories like copyright/licence info (#1010).
- `content`: a `TextField` with the markdown contents.

The contents of these documentation files isn't that big, so I don't think it's an issue to store it directly. In a form, it also makes more sense to edit the contents directly in the interface, rather than download/upload a file.

Using a single `CorpusDocument` model, rather than a separate field for each type of document, also makes the system more flexible for offering different types of documentation.

### Validation

If you're creating a corpus through the interface, it makes sense that the image and documentation may be added at the very end of that process, since they're more the "fluff". The image and documentation should not be required in the database, but having an image should be a requirement for `corpus.ready_to_publish()`.

PR #1526, opened 2024-04-03
Feature/corpus image saving

lukavdplas:
Changes the way corpus images are saved in the database to an `ImageField`. Part of https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1524

Image files are saved in the media storage (MEDIA_ROOT) for the project. For database-only corpora, this will enable convenient uploading. For Python corpora, the file is copied from the source directory into media storage.

Corpus images are now optional. If no image has been added to a corpus, the application will use a default image that can be configured in the project settings.


backend/addcorpus/migrations/0005_add_validators.py
```diff
@@ -4,6 +4,11 @@
 from django.db import migrations, models
 
 
+def validate_image_filename_extension(filename):
+    allowed = ['.jpeg', '.jpg', '.png', '.JPG']
+    addcorpus.validation.creation.validate_filename_extension(filename, allowed)
```
JeltevanBoheemen, line 9: I would move this to `addcorpus.validation` somewhere, hard to find in migration files.

JeltevanBoheemen: Also, using MIMETYPE is a better way to validate file type, but this will almost certainly work fine.

backend/addcorpus/models.py
```diff
@@ -7,7 +7,7 @@
 
 from addcorpus.constants import CATEGORIES, MappingType, VisualizationType
 from addcorpus.validation.creation import validate_language_code, \
```
JeltevanBoheemen, line 9: I hate this import style, would rather use parentheses. Tiny hill.

backend/addcorpus/validation/publishing.py
```diff
@@ -2,6 +2,7 @@
 This module defines functions to check if a corpus is ready to be published.
 '''
 
+import os
```
JeltevanBoheemen, line 5: Unused import or previously non-working file? ;)

backend/conftest.py
```diff
@@ -111,3 +113,12 @@ def clear_test_corpus(es_client, corpus_name):
     # check existence in case teardown is executed more than once
     if es_client.indices.exists(index = index):
         es_client.indices.delete(index = index)
+
+@pytest.fixture(scope='session', autouse=True)
+def clean_test_data():
+    if os.path.isdir(MEDIA_ROOT):
```
JeltevanBoheemen, line 119: Now troll a developer and make them set `MEDIA_ROOT=/`

JeltevanBoheemen (approved):
The default image is a bit depressing..
Rest is good and clean as always!

backend/addcorpus/migrations/0005_add_validators.py
lukavdplas: I moved it to the migration files because it's not used anymore; ImageField does this validation out-of-the-box. It's still referenced in this old migration, though.

backend/addcorpus/images/default.png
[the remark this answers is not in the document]
lukavdplas: If anyone asks, you can say it's to motivate corpus creators to set a custom picture.

PR #1527, opened 2024-04-04
Add sentry to django backend

fliepeltje:
Adds `sentry_sdk` config


backend/ianalyzer/settings.py
```diff
@@ -20,6 +21,9 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = 'kxreeb3bds$oibo7ex#f3bi5r+d(1x5zljo-#ms=i2%ih-!pvn'
 
+if SENTRY_DSN := os.getenv('SENTRY_DSN'):
+    sentry_sdk.init(dsn=SENTRY_DSN, tracing=True)
```
BeritJanssen, line 25: I cannot find the `tracing` option in the [documentation](https://docs.sentry.io/platforms/python/configuration/options/). What does it do? If this means we'll go as deep as tracing SQL queries, I wouldn't think this is ncessary.

BeritJanssen (changes requested):
Thank you! I would want to know that building works with the updated requirements. The Github Actions only run on branches prefixed with `feature/`, so cannot get this information from the jobs we configured right now.

In order to use Sentry in production we'll also need to add this information to the deployment module's settings. Or we move the Sentry initialization to `ianalyzer.common_settings`, but then it would still be good to be able to override `profiles_sample_rate` and `traces_sample_rate`, because I do think in production we don't want to flood Sentry with hundreds of repetitions of the same bug. Therefore, it would be nice if the values used here (i.e., 1.0) would be explicit here, to make sure that we see which values are now active.

backend/ianalyzer/settings.py
lukavdplas, line 1: The way I-analyzer is set up, this `settings.py` module is _exclusively_ used in development servers, so adding settings here makes sense if you _don't_ want them to show up in production.

If sentry should be supported in all development and production servers, it should be added in `common_settings.py` instead. If it's intended for production only, it should be added in the deployment repository. (In that case, you could still enable sentry on a specific development server by adding it to your `settings_local.py`.)

PR #1528, opened 2024-04-04
Add sentry to django backend

fliepeltje:


backend/ianalyzer/common_settings.py
```diff
@@ -4,10 +4,19 @@
...
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+if SENTRY_DSN := os.getenv('SENTRY_DSN'):
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        enable_tracing=True,
+        sample_rate=1.0,
+        traces_sample_rate=1.0
+    )
+
```
BeritJanssen, line 19: Sorry for yet another change request, but will this mean we cannot override the values that are set here? Perhaps the best place for this would be in `backend.manage.main()`, right after the `os.environ.setdefault...`. Then the `common_settings` would have:
```
SENTRY_DSN := os.getenv('SENTRY_DSN')
SENTRY_ENABLE_TRACING=True
SENTRY_SAMPLE_RATE = 1.0
SENTRY_TRACES_SAMPLE_RATE = 1.0
```

And `manage.py`:
```
if SENTRY_DSN:
   sentry_sdk.init(
      dsn=SENTRY_DSN,
      enable_tracing=SENTRY_ENABLE_TRACING,
      sample_rate=SENTRY_SAMPLE_RATE,
      traces_sample_rate=SENTRY_TRACES_SAMPLE_RATE
)
```
That way, we would be able to set different values for `enable_tracing` etc. per instance, but keep a Sentry init for all instances.

BeritJanssen (changes requested):

backend/ianalyzer/settings.py
```diff
@@ -20,6 +20,8 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = 'kxreeb3bds$oibo7ex#f3bi5r+d(1x5zljo-#ms=i2%ih-!pvn'
 
+
+
```
BeritJanssen, line 24: No need to add emtpy lines here.

BeritJanssen:
I added a branch in the [deployment module](https://github.com/UUDigitalHumanitieslab/deployment/tree/sentry) which should make it easier to add Sentry to the different RSLab applications. So it's not necessary to include this in the code base of the individual repos.

Issue #1347, opened 2023-12-05
Adjusting visualisation options scrolls view


lukavdplas:
### What went wrong?

Adjusting options for the barchart (e.g. normalisation) also scrolls the view to the top, which is very annoying.

There is enough clutter above the visualisations that, on a full HD screen with a windows taskbar, you'll probably need to scroll down in order to have the visualisation completely in view. It's weird that adjusting options scrolls the window away from the visualisation.

### What did you expect to happen?

The scroll position should not change when adjusting visualisation options.

### Screenshot

[Screencast where I adjust one of the options for the visualisation, and the window scrolls up immediately](https://github.com/UUDigitalHumanitieslab/I-analyzer/assets/43678097/d05bd109-3e3a-4ffd-b63c-7f3f29f75dc0)

### Where did you find the bug?

- https://ianalyzer.hum.uu.nl

### Version

5.3.0

### Steps to reproduce

- Go to https://ianalyzer.hum.uu.nl/search/troonredes?visualize=resultscount&visualizedField=monarch
- Scroll down a bit
- Adjust the normalisation

Issue #1408, opened 2024-02-01
Add map visualization


BeritJanssen:
**Is your feature request related to a problem? Please describe.**
If a corpus has geographical metadata, we would like to visualize this on a map.

**Describe the solution you'd like**
Perhaps this [ChartJS plugin](https://www.npmjs.com/package/chartjs-chart-geo) is interesting

**Describe alternatives you've considered**
Other geo visualization libraries exist, e.g. maplibre (cf. [in this repo](https://github.com/UUDigitalHumanitieslab/jewish-settlement-inscriptions/blob/develop/src/index.js)

lukavdplas:
Can we add more information to this issue?

Things like:
- What question is the map supposed to answer?
- Where should it be displayed?
- What geographical information should be visible, if any?

lukavdplas:
Some thoughts:

- I already mentioned this in our last meeting, but for the record: if we intend to work on https://github.com/UUDigitalHumanitieslab/I-analyzer/discussions/1427 in the future and roughly follow the implementation suggested there (both are big ifs), it would be a good idea to build new visualisations with the following principles:
	- Use a visualisation library with _declarative_ syntax, such as Plotly or Vega. Both of those have some support for maps, but I'm not sure if they would meet our needs.
	- Try to implement functionality in the backend rather than the frontend. Centre the visualisation around a _backend_ endpoint that essentially takes the query and some additional parameters and returns the result - rather than using the _frontend_ to construct aggregation queries that the backend merely forwards to elasticsearch.
- For the frontend component, `Results` is a useful base class for fetching data based on a query, but it hasn't yet been implemented in other visualisations (https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1468 ).
- I added a "needs mockup" label to the issue because it's the kind of thing that could benefit from creating a picture before you get into the nitty gritty, but I'm not involved in this project, so it's up to you.

ar-jan:
The initial map component has been merged into develop, the actual map visualization is to be implemented (probably requires Angular to be updated).

ar-jan:
#1549 works with ngx-maplibre-gl, but it's worth checking if the features we need are supported by Vega or Plotly.

Vega-Lite (and therefore Vega-Altair which produces Vega-Lite) does not yet support zoomable maps.

If we do want to use maplibre, we need to add a different base layer (include local assets, or use a service?) and fix the tests.

lukavdplas:
[...]

PR #1529, opened 2024-04-04
Feature/angular 17

### LINKED ISSUE #1329: update angular
[documentation for primeng v14](https://www.primefaces.org/primeng-v14-lts/) (the version we use) shows a moving, undismissable banner about how we're not licensed to use it anymore - or at least the language is intended to suggest that. Anyway, we should probably do something about it - if nothing else, it does achieve the desired effect of making the documentation hostile to use.

The quick solution is to upgrade angular and primeng (which depends on angular). Though I would personally support moving toward a component library that doesn't pull stunts like this.

JeltevanBoheemen:
Resolves #1329 

A few important notes:

- Requires Node 18, provided a `.nvmrc` file with the server version
- Reopens #1347. Working on an Ivy distribution of the package to fix it again
- Big version jump for `ng2-pdf-viewer`. i don't have data to test this. @lukavdplas or @BeritJanssen would you be so kind to check if the pdf viewer still works?


.nvmrc
```diff
@@ -0,0 +1 @@
+18.17.1
```
ar-jan, line 1: Maybe just `18`?

ar-jan (approved):
Looks good to me. I've only tested locally with the JMIG corpus, there everything is functional.

You can address this console warning `defaultLabel property is deprecated since 16.6.0, use placeholder instead` for primeng multiselect.

Note for Docker use, need to rebuild the containers or webpack complains about Edge version.

lukavdplas:
> Reopens https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1347. Working on an Ivy distribution of the package to fix it again

That bug can cause some very annoying behaviour (if I recall correctly, it was not limited to visualisation options); since this update does not bring any direct improvements to the user, can we hold off on merging it?

> Big version jump for ng2-pdf-viewer. i don't have data to test this. @lukavdplas or @BeritJanssen would you be so kind to check if the pdf viewer still works?

I don't have any PDF data locally; perhaps the test server is the easiest place to check this?

.nvmrc
JeltevanBoheemen: There are a few libraries that are pinned lower than they could be because they require a higher minor verion of node. So the specific pin to `18.17.1` is necessary here.

BeritJanssen:
#1408 depends on a newer Angular version, so I'm afraid we cannot hold this off for too long.

ar-jan:
By the way, can these warnings be ignored or should they be addressed?

> warning " > ng2-pdf-viewer@10.0.0" has unmet peer dependency "pdfjs-dist@~2.16.105".
> warning " > terser-webpack-plugin@5.3.6" has unmet peer dependency "webpack@^5.1.0".

JeltevanBoheemen:
> By the way, can these warnings be ignored or should they be addressed?
> 
> > warning " > ng2-pdf-viewer@10.0.0" has unmet peer dependency "pdfjs-dist@~2.16.105".
> > warning " > terser-webpack-plugin@5.3.6" has unmet peer dependency "webpack@^5.1.0".

I have added them to the peer dependencies in `package.json`. Oddly, when doing things directly in the frontend (such as updating browserslists), the warnings still occur. Anyway, safe to ignore.

Issue #1256, opened 2023-09-11
dutch annual reports: word model documentation does not load


lukavdplas:
Trying to open the word model documentation of Dutch Annual Reports on the server gives a never-ending loading spinner. The Troonredes documentation loads fine, though, so perhaps the corpus is not set configured correctly.

PR #1531, opened 2024-04-05
Feature/corpus document model

### LINKED ISSUE #1010: License / copyright information per corpus
In the interest of FAIR data, it would be nice if users could see what they are allowed to with downloaded csv data. I don't think any of our corpora specify this in their documentation at the moment.

We could add such information in the documentation file for the corpus. If we want to use a full license, it would make more sense to make that a separate page.

### LINKED ISSUE #1524: Change data model for corpus attachments

lukavdplas:
Updates the database model and API for corpus documentation. (close #1524)

The contents of documentation pages are imported into the database, which will allow us to support documentation for database-only corpora.

Instead of implementing each type of documentation separately, there is now a single `CorpusDocumentationPage` model. The API is updated so the frontend requests all documentation pages for a corpus in a single endpoint, rather than using an endpoint for each page.

Other changes:
- Corpora now support two new types of documentation: licence and terms of service (close #1010).
- Templating is now supported on _all_ documentation pages, not just the citation guidelines. (Not really necessary, it was just simpler to make this generally available than to write an exception for citations.)
- Documentation pages can be viewed and edited in the admin (though like the rest of the corpus configuration, changes are reset when the corresponding Python corpus definition is imported).
- The default expectations for the corpus directory structure (e.g. general documentation should be in a folder named "description") are still the same, but you can now override that per corpus.
- Fixes the never-ending loading spinner described in #1256.


backend/addcorpus/models.py
```diff
@@ -408,3 +397,35 @@ def clean(self):
                     'the corpus or correct the following errors.',
                     e
                 ])
+
+class CorpusDocumentationPage(models.Model):
+    class PageType(models.TextChoices):
```
JeltevanBoheemen, line 402: I didn't know this base class existed for choices. Much easier to read then the nested tuple business.

backend/addcorpus/models.py
```diff
@@ -408,3 +397,35 @@ def clean(self):
...
+    content = models.TextField(
+        help_text='markdown contents of the documentation'
+    )
+
+    class Meta:
+        constraints = [
+            UniqueConstraint(
+                fields=['corpus_configuration', 'type'],
+                name='unique_documentation_type_for_corpus'
+            )
```
JeltevanBoheemen, line 430: clean.

documentation/Corpus-documentation.md
```diff
@@ -0,0 +1,19 @@
+# Corpus documentation pages
+
+Corpora can include markdown templates containing documentation. I-analyzer supports several types of documentation pages.
+
+- general information
+- citation guidelines
+- documentation on word models
+- a licence for re-using the data
+- terms of service (usually pro)
```
JeltevanBoheemen, line 9: Define `usually pro` ?

documentation/Corpus-documentation.md
```diff
@@ -0,0 +1,19 @@
...
+
+All are optional.
+
+### Writing documentation
+
+Documentation files are written in markdown. For information about supported flavours, see [marked.js documentation](https://marked.js.org/) - the frontend uses that package with default options.
```
JeltevanBoheemen, line 15: Maybe refer to an existing documentation page in the source code for inspiration.

JeltevanBoheemen (approved):
Really important change for the scientific credibility of the application, nicely done.
I'm 50/50 on supporting different paths to documentation, I would be fine with enforcing the default values. But it's opt-in so it's okay.

documentation/Corpus-documentation.md
lukavdplas: ah, I think I just walked away from the computer halfway through that sentence.... 😅

PR #1534, opened 2024-04-10
increment version number on CITATION.cff

BeritJanssen:
To keep version numbers up to date in `CITATION.cff`, I thought to update upon every merge to `master`. This may only a partial solution for keeping `CITATION.cff` up to date, as the doi from zenodo will still point to the old version. *BUT* we may use the overarching DOI for I-Analyzer, [10.5281/zenodo.10261620](https://zenodo.org/account/settings/github/repository/UUDigitalHumanitieslab/I-analyzer), which seems to point to the latest version always.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,21 @@
+name: Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'master'
+  jobs:
+    citation-update:
+    uses: actions/checkout@v3
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Update CITATION.cff
+    - run: |
+      version=`grep -o '\d\+\.\d\+\.\d\+' package.json`
+      sed "s/^version: [[:digit:]]\{1,\}\.[[:digit:]]\{1,\}\.[[:digit:]]\{1,\}/version: $version/" CITATION.cff > CITATION.cff.temp
+      rm -f CITATION.cff
+      mv CITATION.cff.temp CITATION.cff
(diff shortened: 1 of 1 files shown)


lukavdplas:
Is the motivation of this workflow that we won't need to update the citation file manually?

BeritJanssen:
Yes, precisely.

BeritJanssen:
Ah, good points. Triggering the workflow on push to `release` branches will also mean we'll integrate the changed citation file in develop and main/master both. Will update that & also automate the date. Should I additionally change `CITATION.cff` to point to the top-level I-Analyzer doi, rather than the latest version? I can also update the manual's link to that page (which lists all the versions). Now the manual points to an outdated version, indeed.

BeritJanssen:
Turns out the doi in the CITATION file was already correct, the badge on our repo page is too. So only adjusted the link in the documentation.

lukavdplas:
Neat!

Two more (small) comments:

- The release checklist should mention that you now _need_ to push the release branch to github.
- What happens if the citation/date are already up to date? (Most likely because the release branch is pushed multiple times.) There shouldn't be a commit in that case.

lukavdplas:
> I thought that the release or hotfix branch would always be pushed, as that's the one we'd be testing on the test server prior to merging it to master & develop. Doesn't git flow release publish, as mentioned in [the documentation](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/documentation/Making-a-release.md#check-if-everything-works) do a push?

True, but that section of documentation also suggests you publish the branch for one specific purpose. If there is another reason to publish the branch, it doesn't hurt to mention it, I think.

PR #1539, opened 2024-04-11
feat: introduce authService as dependency on tagService

### LINKED ISSUE #1483: Hide tag features for unauthenticated users
As we implement #988 , we want to hide the following from unauthenticated users:
- TagFilter
- Tag control in DocumentView

Under the hood, a `FoundDocument` fetches the tags connected to a document, which we would want to avoid (as it predictably fails) for unauthenticated users.

`QueryModel` and `FoundDocument` need to be adjusted such that the logic related to tags is switched on or off, depending on whether or not we are dealing with an authenticated user.

Rewrite I-Analyzer from scratch (not really though)

BeritJanssen:
Close #1483. After 4+ hours of trying to write a decent unit test for `document-view-component` which tests that tags are shown for authenticated users / not shown for unauthenticated users, I give up, and commit with a test only for the behaviour of the tag filter. 😞 Hope to figure out what the problem is soon. I think it's something to do with the `TagServiceMock` not mocking convincingly enough. But in the interest of a speedy release, perhaps better to just rely on my manual tests for now.

### Authorized:
Screenshot 2024-04-11 at 15 56 34
Screenshot 2024-04-11 at 15 56 39

### Unauthorized:

Screenshot 2024-04-11 at 15 56 50
Screenshot 2024-04-11 at 15 56 54

### Requests for unauthenticated user:
The only request to the backend is now one to the `user` endpoint:

Screenshot 2024-04-11 at 15 57 30

### DIFF
--- a/frontend/src/app/document-view/document-view.component.html
+++ b/frontend/src/app/document-view/document-view.component.html
@@ -9,10 +9,10 @@
                             <ia-search-relevance [value]="document.relevance"></ia-search-relevance>
                         </td>
                     </tr>
-                    <tr>
+                    <tr *ngIf="(document?.tags$ | async) as tags">
                         <th>Your tags</th>
                         <td>
-                            <ia-document-tags [document]="document"></ia-document-tags>
+                            <ia-document-tags [document]="document" [tags]="tags"></ia-document-tags>
                         </td>
                     </tr>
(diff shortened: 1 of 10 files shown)


JeltevanBoheemen (approved):
This looks good!
Props for trying to get the test working, no problem that it didn't work out.

PR #1541, opened 2024-04-11
Feature/corpus readers

### LINKED ISSUE #981: Import/export corpus definitions

### LINKED ISSUE #985: Index corpus from interface
Allow curators to run indexing commands from the corpus definition interface (#982)

JeltevanBoheemen:
- Provides utility reader for json corpora, partially resolves #985 
- Detects if corpus exists, does not override if this is the case, partially resolves #981


backend/addcorpus/reader.py
```diff
@@ -0,0 +1,38 @@
...
+
+def make_reader(corpus: Corpus, data_directory: str = None) -> Reader:
+    '''
+    From a corpus, returns a Reader object that allows source extraction
+
+    For Python corpora, simply loads the definition class,
+    for JSON based corpora, construct Reader from the database.
+    '''
+    if corpus.has_python_definition:
+        return load_corpus_definition(corpus.name)
```
lukavdplas, line 26: It's a bit inconsistent that `data_directory` is required for corpora without a Python definition, but completely ignored without one. Especially since you could otherwise call `make_reader` without caring about whether the corpus has a Python definition or not.

Perhaps we can fill in the directory for `NewReader` too? For example

```py
data_directory = os.path.join(settings.SOURCE_DATA_ROOT, corpus.name)
```

or we can add that to `CorpusConfiguration`, so this function can just use

```py
data_directory = corpus.data_directory
```

In the future, we will still need a way to customise the directory for a corpus, but this should work for now.

backend/addcorpus/tests/test_reader.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+from addcorpus.models import Corpus
+from addcorpus.reader import make_reader
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+
+
+def test_make_reader_python(mock_corpus):
+    py_corpus = Corpus.objects.get(name=mock_corpus)
+    reader = make_reader(py_corpus)
```
lukavdplas, line 11: This is a bit confusing since `py_corpus` is not the Python definition.

```suggestion
    corpus = Corpus.objects.get(name=mock_corpus)
    reader = make_reader(corpus)
```

lukavdplas (changes requested):
A few minor points after going over this again.

PR #1546, opened 2024-04-16
Feature/optional bounds for filters

### LINKED ISSUE #1535: Infer upper/lower for RangeFilter and DateFilter
RangeFilter and DateFilter **require** an upper and lower. These are not supplied in JSON corpora.

Those should be inferred after indexing, using an ES request to determine the values.

An alternative would be to supply these manually in the corpus definition. This requires more work for the user.

lukavdplas:
This makes the specification of `upper` and `lower` bound for the range filter and date filter optional (close #1535). If no upper or lower bound has been defined, the frontend will request the min/max values when rendering the filter component.

Summary of changes to the code:
- Refactored the methods for handling aggregations in the `searchService`. `aggregateSearch()` is now as general as the name implies, and allows aggregations other than the `terms` type. `dateHistogramSearch()` is removed.
- Added an abstract `Aggregator` class with subclasses for specific aggregation types. This class handles the translation to an elasticsearch query, and translation from an elasticsearch result.
- Type definitions treated the results of `terms` and `date_histogram` aggregations as the same type, but they are not compatible. Changed the definitions so these are different return types.
-  The `RangeFilter` and `DateFilter` now accept `undefined` min/max values (i.e. no bound). If the corpus did not specify an upper/lower bound, these will be the initial values.
- When the filter widget for the range/date filter is rendered, it will add a a finite upper/lower bound to the filter, if it doesn't have one already. This is done with an aggregation query.


frontend/src/app/filter/date-filter/date-filter.component.ts
```diff
@@ -17,16 +19,50 @@ export class DateFilterComponent extends BaseFilterComponent<DateFilter> {
...
+        }
+        return this.fetchAggregation(new MaxDateAggregator(filter.corpusField));
+    }
+
+    private fetchAggregation<Result>(aggregator: Aggregator<Result>): Promise<Result> {
+        const queryModel = new QueryModel(this.queryModel.corpus);
+        return this.searchService.aggregateSearch(
+            queryModel.corpus, queryModel, aggregator
+        );
+    }
```
JeltevanBoheemen, line 67: clean.

frontend/src/app/filter/multiple-choice-filter/multiple-choice-filter.component.ts
```diff
@@ -29,17 +30,15 @@ export class MultipleChoiceFilterComponent extends BaseFilterComponent<MultipleC
...
-                        o => o.label
-                    )
-                ).catch(() => this.options = []);
-
+            this.searchService.aggregateSearch(queryModel.corpus, queryModel, aggregator).then(result =>
+                this.options = _.sortBy(
+                    result.map(x => ({ label: x.key, value: x.key, doc_count: x.doc_count })),
+                    o => o.label
+                )
+            ).catch(() => this.options = []);
```
JeltevanBoheemen, line 41: I would prefer meaningful variable names over `x` and `o`, or type annotations. The code is already quite difficult to read.

I know this is not in the scope of the PR, but just stood out to me now that it's presented to me :)

frontend/src/app/filter/range-filter/range-filter.component.html
```diff
@@ -1,6 +1,8 @@
-<div class="range" *ngIf="(filter?.state$ | async).data as data">
-    <span>{{data.min}} - {{data.max}}</span>
-    <p-slider [animate]="true" [ngModel]="[data.min, data.max]"
-        [range]="true" [min]="min" [max]="max"
-        (onChange)="sliderValue$.next($event.values)"></p-slider>
-</div>
+<ng-container *ngIf="min !== undefined && max !== undefined">
```
JeltevanBoheemen, line 1: I presume the `*ngIf` is to delay loading the component until after the `min` and `max` are set?
In that case we need to be sure that these are always set.

frontend/src/app/models/aggregation.ts
```diff
@@ -0,0 +1,163 @@
+import { CorpusField } from './corpus';
```
JeltevanBoheemen, line 1: By god this is great.

JeltevanBoheemen (approved):
I have to admit it's a bit of an end-of-the-day review, but the test coverage gives me the confidence to approve it anyway.

I especially like the `models/aggregation.ts` file!

frontend/src/app/models/aggregation.ts
JeltevanBoheemen: Not the import. The file.

frontend/src/app/filter/range-filter/range-filter.component.html
lukavdplas: Yes, you can't really make a slider otherwise. The same component _should_ be requesting a min/max value.

PR #1547, opened 2024-04-16
corpus JSON export

### LINKED ISSUE #1520: Add `url`  display type
This affects places where the frontend renders the content of fields:

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/fb1d88c068bc7e8a74372baa0c6951c1eba90c14/frontend/src/app/document-view/document-view.component.html#L21-L29

There is a special rule for URL fields (rendering the content as a link), but this is based on the field name.

It would be more appropriate to set te the `display_type` attribute accordingly.

Suggestion:
- Expand the options for `display_type` to include `'url'`.
- Adjust corpus definitions to use the url display type.
- Adjust the document view component and document preview component so they use the `display_type` instead.

lukavdplas:
- Adds a function to convert a `Corpus` object to a JSON. (part of #981)
- Adds `url` display type for fields (close #1520). Displaying fields as links is now based on the `display_type` rather than the name. (This was required for a clean export of URL fields.)


JeltevanBoheemen:
I won't get around to reviewing this before next week!

backend/addcorpus/json_corpora/export_json.py
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+def export_date(date: date):
+    return date.strftime(DATE_FORMAT)
+
+def export_corpus_source_data(configuration: CorpusConfiguration) -> Dict:
+    data = {
+        'type': 'csv'
+    }
+    if configuration.source_data_delimiter != DEFAULT_CSV_DELIMITER:
```
JeltevanBoheemen, line 39: I believe `configuration.source_data_delimiter` has a default value right? In that case the check can be skipped and just set to that value.

frontend/src/app/document-view/document-view.component.ts
```diff
@@ -68,7 +68,7 @@ export class DocumentViewComponent implements OnChanges {
     }
 
     isUrlField(field: CorpusField) {
-        return field.name === 'url' || field.name.startsWith('url_');
+        return field.displayType === 'url';
```
JeltevanBoheemen, line 71: This is unrelated to this PR?

frontend/src/app/models/corpus.ts
```diff
@@ -64,7 +64,8 @@ export interface DocumentContext {
 }
 
 
-export type FieldDisplayType = 'text_content' | 'px' | 'keyword' | 'integer' | 'text' | 'date' | 'boolean';
+export type FieldDisplayType =
+    'text_content' | 'px' | 'keyword' | 'integer' | 'text' | 'date' | 'boolean' | 'url';
```
JeltevanBoheemen, line 68: Same as above.

JeltevanBoheemen (approved):

frontend/src/app/document-view/document-view.component.ts
lukavdplas: It's somewhat related. The URL display type was needed for the export because the JSON schema has that option, but the database did not.

backend/addcorpus/json_corpora/export_json.py
lukavdplas: Yes, but if you are using the default value, you don't need to specify the delimiter in the JSON at all. So this check is needed if importing + exporting should return the same JSON.

backend/addcorpus/json_corpora/export_json.py
```diff
@@ -0,0 +1,95 @@
+from typing import Dict
+from datetime import date
+from addcorpus.models import Corpus, CorpusConfiguration, Field
+from addcorpus.json_corpora.constants import DEFAULT_CSV_DELIMITER, DATE_FORMAT
+from addcorpus.es_mappings import primary_mapping_type
+
+def export_json_corpus(corpus: Corpus) -> Dict:
+    config = corpus.configuration
+    data = {'name': corpus.name}
+    data['meta'] = export_corpus_meta(config)
```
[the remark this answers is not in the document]
lukavdplas: Hm, the same goes for the `import_json_corpus` function (which this one mirrors). Maybe just `corpus_to_json` / `json_to_corpus` ?

PR #1548, opened 2024-04-17
Feature/more citation docs

### LINKED ISSUE #1322: Citation documentation per corpus

lukavdplas:
Adds citation documentation for the Times and Dutch Newspapers Public corpora.

close #1322

### DIFF
--- a/backend/corpora/dutchnewspapers/citation/citation_public.md
+++ b/backend/corpora/dutchnewspapers/citation/citation_public.md
@@ -1,0 +1,43 @@
+If you cite this corpus, we recommend the following format.
+
+## Citing the whole corpus
+
+To cite the entire corpus, follow citation guidelines for databases or datasets.
+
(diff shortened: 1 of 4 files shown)


tijmenbaarda:
This is excellent! It gives examples but it also explains very well why a certain way of citing is recommended.

I have not tested whether this is working correctly from a technical perspective, but I assume it does.

tijmenbaarda (approved):

PR #1549, opened 2024-04-24
Map visualization

ar-jan:
WiP on top of #1529.

Includes update to rxjs 7, needed for ngx-maplibre-gl.

I couldn't get ngx-maplibre-gl to work without setting `allowSyntheticDefaultImports`, although it seems like this shouldn't be needed.

TODO:

- [ ] Use different style / source tiles for the base layer
- [ ] Labels? (link to record?)
- [ ] Clustering?


ar-jan:
Map visualization was instead added using vega in #1629.

PR #1555, opened 2024-05-02
Instructions to run loadcorpora admin command

tijmenbaarda:
As discussed with @BeritJanssen the READMEs did not include instructions to run the `loadcorpora` admin command before indexing, so I added it to the readmes.

I have also done a small fix. The documentation states that the keys in the `CORPORA` dictionary are case insensitive. However, the function `load_corpus_definition` does not match using lower case on both sides, so that using lower case is obligatory. I think this small change fixes this, but please check it carefully.

I'm asking a review from @lukavdplas because they have been most involved with documentation.


lukavdplas (approved):
:+1:

PR #1556, opened 2024-05-03
Update XML corpus definitions

lukavdplas:
https://github.com/UUDigitalHumanitieslab/ianalyzer-readers/pull/18 proposes an update to the `XMLReader` that includes breaking changes. This PR migrates definitions for corpora with XML data to the new interface.

I'm just opening the draft PR to illustrate how that update would affect i-analyzer. This version pulls the (unreleased) changes in `ianalyzer_readers` by pinning the commit hash, but we shouldn't merge this until a new release is available on pip.


backend/corpora/dbnl/dbnl.py
```diff
@@ -261,18 +262,18 @@ def _xml_files(self):
             Pass(
                 Backup(
                     XML( # get the language on chapter-level if available
+                        CurrentTag(),
                         attribute='lang',
                         transform=lambda value: [value] if value else None,
                     ),
                     XML( # look for section-level codes
-                        {'name': 'div', 'attrs': {'type': 'section'}},
+                        Tag('div', type='section'),
```
BeritJanssen, line 270: Can this search for a given attribute be passed as a flat keyword argument, or shouldn't it be `attrs={'type': 'section'}` instead?

lukavdplas: BeautifulSoup supports either. In some cases, `attrs` is preferred to prevent overlap with named arguments to `find_all()`.

backend/corpora/parliament/finland.py
```diff
@@ -17,23 +18,9 @@ def format_role(values):
         clean_id = id.replace('#', '')
         return roles.get(clean_id, clean_id)
 
-def speech_metadata(speech_node):
-    """Gets the `note` sibling to the speech."""
-    return speech_node.find_previous_sibling('note')
+# def find_topic(speech_node):
+#     return speech_node.parent.find_previous_sibling('head')
```
BeritJanssen, line 22: Work in progress? Or why is this commented out?

backend/corpora/parliament/ireland.py
```diff
@@ -6,6 +6,7 @@
 from bs4 import BeautifulSoup
```
BeritJanssen, line 6: Is this import still needed?

backend/corpora/rechtspraak/rechtspraak.py
```diff
@@ -5,6 +5,8 @@
 from os import makedirs, remove
 from typing import Optional
 from zipfile import ZipFile, BadZipFile
+from ianalyzer_readers.xml_tag import Tag, ParentTag
+import bs4
```
BeritJanssen, line 9: what is this bs4 import for?

BeritJanssen (commented):
LGTM, nice work! I do wonder whether the BeautifulSoup imports in the corpus definitions are still needed, perhaps good to check in an editor before merging.

BeritJanssen:
I made a release 0.2.0 of ianalyzer-readers, and updated the requirements accordingly.

backend/corpora/parliament/ireland.py
lukavdplas: Yes, it's used to parse metadata.

backend/corpora/dutchnewspapers/dutchnewspapers_public.py
```diff
@@ -223,19 +213,11 @@ def fields(self):
...
-                                  },
-                                  external_file={
-                                      'xml_tag_toplevel': 'DIDL',
-                                      'xml_tag_entry': 'dcx'
-                                  }
-                                  ),
+            extractor=XML(
+                lambda metadata: Tag('recordIdentifier', string=metadata['id']),
+                SiblingTag('subject'),
+                external_file=True
```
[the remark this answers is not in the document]
lukavdplas: I mean, I agree that a complex yet untested corpus like this is broken by design, but this is already a huge PR so I would rather not stretch the scope any further.

Since an update to the corpus has been in progress for quite some time now, I had assumed that that process includes writing some unit tests _before_ starting a huge index job. I didn't want to spend too much time on this corpus as I assumed it would lead to double work.

PR #1567, opened 2024-05-06
Feature/index from db

### LINKED ISSUE #1542: Index corpus without Python definition
Make it possible to index a corpus that does not have a Python definition.

Tasks:
- [x] enable database access during indexing: #1533
- [x] enable source data extraction without a Python corpus: #1541 
- [ ] rewrite index functions to be based on the database model

JeltevanBoheemen:
Resolves #1542

Some changes unrelated to issue to satisfy linter. Sorry..


backend/corpora/jewishmigration/test_jewishmigration.py
```diff
@@ -193,15 +209,16 @@ def test_geofield(jm_client, jm_corpus):
...
+    results = jm_client.search(index=es_index, query=query)
     assert results['hits']['total']['value'] == 1
 
 
 def test_data_from_request(jm_corpus, monkeypatch):
     monkeypatch.setattr(requests, "get", mock_get)
-    sources = jm_corpus.sources(
-        start=jm_corpus.min_date, end=jm_corpus.max_date)
-    documents = list(jm_corpus.documents(sources))
+    corpus_def = load_corpus_definition(jm_corpus.name)
```
BeritJanssen, line 218: Shouldn't this also do `Corpus.objects.get`, like the `es_index` script?

backend/es/tests/test_alias.py
```diff
@@ -1,38 +1,30 @@
...
-    res = es_alias_client.indices.get_alias(name=corpus_definition.es_index)
+def test_alias(db, es_alias_client):
+    corpus = Corpus.objects.get(name='times')
+    assert corpus.configuration.es_index == 'times-test'
+    alias(corpus)  # create an alias ianalyzer-times
+    res = es_alias_client.indices.get_alias(name=corpus.configuration.es_index)
     assert res.get('times-test-2') is not None
 
+
 def test_alias_with_clean(es_alias_client):
```
BeritJanssen, line 13: shouldn't this test also have the `db` fixture? Then again, if it doesn't fail...

BeritJanssen (approved):
Noice!

lukavdplas:
Note: I cancelled the backend test job because it seemed to be stuck. I can reproduce the issue on my local machine: one of the tests in `test_es_index` seems to go on forever. I'll try to fix the issue.

lukavdplas:
Update: the issue was caused by `glob` receiving `''` as the directory; it doesn't seem to handle that well. It's fixed now (by adding some extra validation.)

PR #1568, opened 2024-05-06
Bugfix/wordcloud

### LINKED ISSUE #1558: DBNL: content worcloud returns error
### What went wrong?

In the DBNL corpus, the wordcloud ("most frequent words") visualisation does not work for the "content" field. It _does_ work for the title and chapter fields, however.

If you select the content field, the visualisation will show a 500 error message.

A more minor bug: if you switch back to the title field after seeing the error message, the visualisation will not load (it will just show a loading spinner).

- The wordcloud for the content field should work without errors
- *If* an error occurs for one field, the visualisation should gracefully recover when you switch to a field that does work. It looks like some error isn't being reset properly.

### Screenshot

screenshot of I-analyzer showing error message

- [ ] a local server

### Version

5.6.1

### Steps to reproduce

- Go to https://ianalyzer.hum.uu.nl/search/dbnl?query=test&visualize=wordcloud&visualizedField=content to see the error message
- Switch to the "title" field; you will only see a loading spinner. You have to reload the page to see the title wordcloud.

### LINKED ISSUE #1559: Wordcloud table does not show
### What went wrong?

If you switch to the table view for the wordcloud (most frequent words) visualisation, the table is not shown.

This view should show a table with the results.

### Screenshot

screenshot of i-analyzer

- [ ] a local server

### Version

5.6.1

### Steps to reproduce

- Go to https://ianalyzer.hum.uu.nl/search/dbnl?query=test&visualize=wordcloud&visualizedField=title
- Select the table view

lukavdplas:
Bugfixes for the wordcloud visualisation.

close #1559
close #1558

### DIFF
--- a/backend/visualization/wordcloud.py
+++ b/backend/visualization/wordcloud.py
@@ -8,7 +8,7 @@ from es import download as download
 def field_stopwords(corpus_name, field_name):
     corpus = Corpus.objects.get(name=corpus_name)
     field = corpus.configuration.fields.get(name=field_name)
-    if field.language and field.language is not 'dynamic':
+    if field.language and field.language != 'dynamic':
         return get_nltk_stopwords(field.language)
     else:
         return []
--- a/frontend/src/app/visualization/wordcloud/wordcloud.component.html
+++ b/frontend/src/app/visualization/wordcloud/wordcloud.component.html
@@ -4,7 +4,7 @@
     </div>
 
     <div class="block" *ngIf="asTable">
-        <ia-freqtable [headers]="tableHeaders" [data]="results.result$ | async" [defaultSort]="'doc_count'" [name]="(result.state$ | async).field?.name">
+        <ia-freqtable [headers]="tableHeaders" [data]="results.result$ | async" [defaultSort]="'doc_count'" [name]="(results.state$ | async).field?.name">
         </ia-freqtable>
     </div>
 </div>


BeritJanssen (approved):

backend/visualization/wordcloud.py
```diff
@@ -8,7 +8,7 @@
 def field_stopwords(corpus_name, field_name):
     corpus = Corpus.objects.get(name=corpus_name)
     field = corpus.configuration.fields.get(name=field_name)
-    if field.language and field.language is not 'dynamic':
+    if field.language and field.language != 'dynamic':
```
BeritJanssen, line 11: Linters will actually replace `!=` by `is not`. Does `is not` cause a bug here?

lukavdplas: I wasn't really aware of this either, but yes, `is not` checks for identity rather than equality, which caused the issue.

From https://stackoverflow.com/questions/2209755/python-operation-vs-is-not

>  When comparing to `None`, always use is not. None is a singleton in Python - there is only ever one instance of it in memory.
> 
> By comparing identity, this can be performed very quickly. Python checks whether the object you're referring to has the same memory address as the global None object - a very, very fast comparison of two numbers.
> 
> By comparing equality, Python has to look up whether your object has an __eq__ method. If it does not, it examines each superclass looking for an __eq__ method. If it finds one, Python calls it. This is especially bad if the __eq__ method is slow and doesn't immediately return when it notices that the other object is None.
> 
> Did you not implement __eq__? Then Python will probably find the __eq__ method on object and use that instead - which just checks for object identity anyway.
> 
> When comparing most other things in Python, you will be using **!=.**

Issue #726, opened 2022-05-20
Include citation info with documents


lukavdplas:
Question from Jo Guldi:

> What about including a recommended citation format (or series of formats) for each speech?

I like this a lot! Including citation information is probably relevant for non-parliament corpora as well, but it seems very useful for people who use I-analyzer corpora for research.

PR #1572, opened 2024-05-16
Feature/netherlands citation page

BeritJanssen:
This branch adds a `netherlands.md` citation page. Related to #726, but with the solution proposed in #1322.


lukavdplas (approved):
Great to have this! :)

backend/corpora/parliament/citation/netherlands.md
```diff
@@ -0,0 +1,33 @@
...
+To cite a speech in the *Dutch Parliamentary data* corpus, you can retrieve a link by clicking the *link* icon underneath the speech's document tile. This should give you an url as follows:
+{{ frontend_url }}/document/parliament-netherlands/ParlaMint-NL_2021-12-21-eerstekamer-4.u1
+
+### APA style
+
+> Rutte, M. (2021). In *Report of the meeting of the Dutch Lower House, Meeting 37, Session 2 (2021-12-21)*. {{ frontend_url }}/document/parliament-netherlands/ParlaMint-NL_2021-12-21-tweedekamer-2.u225
+
+### MLA style
+
+> Rutte, Mark. *Report of the meeting of the Dutch Lower House, Meeting 37, Session 2 (2021-12-21)*, 2021. {{ frontend_url }}/document/parliament-netherlands/ParlaMint-NL_2021-12-21-tweedekamer-2.u225
```
lukavdplas, line 33: I remember from my time on the project that the "official" English translations for eerste kamer / tweede kamer are house of commons / senate. Did you get this format from somewhere?

BeritJanssen: This description comes from the actual data (debate title).

BeritJanssen:
I just got the comment from the People & Parliament team that APA / MLA aren't terribly Humanities-friendly citation formats. They suggested a format that I think is Chicago (full names, date last). This may actually also be a good point to consider for other corpora. @lukavdplas

lukavdplas:
The way I learned it is that APA is common for linguistics, MLA for literature and Chicago for history and social sciences. My experience with this is quite limited, but my understanding of it is the following:

From my impression of it, MLA is optimised for citing non-academic texts in different media, which is why it's mainly used in literature and cultural analysis. I dislike MLA because it is, imo, good for nothing _but_ cultural analysis, but I think its inclusion is justified and I don't think it's accurate to call it not Humanities-friendly.

APA is optimised for citing academic journals and mainly used in linguistics (and psychology, I would assume), which admittedly is a field not often considered part of the humanities (or only marginally so). However, our own work is defined by the UU categorising it that way. 

I agree Chicago would be a good addition; I left this out in the existing documentation because it would require more of a time investment to familiarise myself with the format.

BeritJanssen:
Thanks for your explanation. I would indeed agree that we shouldn't leave out APA / MLA, because many users of I-Analyzers will be interested in this. I will add Chicago to the list of options in this branch, so then we can hopefully also add it to later `citation.md` files.

PR #1585, opened 2024-05-23
Reorganize and syntax check Github actions

### LINKED ISSUE #1566: Ignore paths for github workflows
The github workflows to run unit tests are run on every push. However, it is not necessary to re-run unit tests the only changes are in `/documentation/*`. Similarly, the backend tests don't need to be re-run if all changes are in `/frontend/*`; vice versa for the frontend tests.

Re-running tests doesn't cause any immediate problems, but it's a waste of time and energy.

**Proposed solution**

Configure `ignore-paths` on the github workflow. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

**Possible complication**

Not sure how this interacts with branch protections. If you make a PR that only updates the documentation, the test workflow would not be run, but we require that these checks pass to merge into develop.

BeritJanssen:
close #1566 : splitting up tests into two .yml files, which was the easiest way to go, and this will automatically ignore changes to README, documentation etc., too. Also fixes an indentation problem in the release workflow.


.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+      - 'develop'
+      - 'master'
+      - 'feature/**'
+      - 'bugfix/**'
+      - 'hotfix/**'
+      - 'release/**'
+      - 'dependabot/**'
+    paths:
+      - 'backend/**'
```
lukavdplas, line 17: This would also ignore changes to `docker-compose.yaml` or the workflow configuration, which could affect the test outcome.

BeritJanssen: I was actually wondering whether the global `package.json` shouldn't also be watched, as the commands in that may change.

lukavdplas: This is why I had suggested `ignore-paths` originally. I think excluding paths is less likely to cause false negatives, especially as the repository is updated over time.

Anyway, a more serious issue is that skipping these checks does not seem to play well with branch protections.

BeritJanssen: I'll look into it. I thought `ignore-paths` was going to get wordier than `paths`, but I'll see how it plays out.

BeritJanssen:
Ah, I can just exclude all `.md` files, so `include-paths` is indeed easier.

PR #1590, opened 2024-05-28
U-Blad Corpus

Meesch:
Adds the U-Blad corpus. This corpus is already indexed and ready to use on the test server! It should not be a lot of work to review this since it is not adding any new functionality. Potentially interesting note is that in this corpus definition the `soup_transform_func` is used not to just extract some text from a node but instead is used to format the strings inside of the node, which could potentially be later expanded with styling features, as hocr contains stylistic classes such as bold/italic etc and font size.


lukavdplas (changes requested):
Nice!

A few fields are missing an appropriate mapping and I'm not sure about the locale setting.

backend/corpora/ublad/description/ublad.md
```diff
@@ -0,0 +1,7 @@
+Op 5 september 1969 kreeg de Universiteit Utrecht voor het eerst een onafhankelijk blad: _U utrechtse universitaire reflexen_. Dit blad kwam voort uit een fusie van twee andere tijdschriften: _Sol Iustitiae_ dat voornamelijk gericht was op studenten en _Solaire Reflexen_ dat meer was bedoeld voor medewerkers. U utrechtse universitaire reflexen was bedoeld voor alle geledingen.
+
+In 1974 veranderde de naam in het _Ublad_. Dat bleef zo tot de universiteit besloot het papieren Ublad digitaal te maken. Onder luid protest verdween het papieren Ublad en ontstond in april 2010 _DUB_, het digitale universiteitsblad.
+
+Om alle informatie uit het verleden toegankelijk te maken, hebben we samen met de Universiteitsbibliotheek de oude jaargangen gescand. In I-analyzer kunt u alle jaargangen van U utrechtse universitaire reflexen en het Ublad vinden en doorzoeken.
```
JeltevanBoheemen, line 5:
```suggestion
Om alle informatie uit het verleden toegankelijk te maken, heeft het Centre for Digital Humanities samen met de Universiteitsbibliotheek de oude jaargangen gedigitaliseerd. In I-analyzer kunt u alle jaargangen van U utrechtse universitaire reflexen en het Ublad vinden en doorzoeken.
```

backend/corpora/ublad/description/ublad.md
```diff
@@ -0,0 +1,7 @@
...
+
+Het onafhankelijke Ublad geeft een kleurrijk verslag van wat er speelde op de universiteit, de stad en het studentenleven door middel van atikelen, foto’s en cartoons. De afbeelding die is gebruikt voor OCR is voor elke pagina bijgevoegd zodat u altijd het originele bronmateriaal kan raadplegen.
```
JeltevanBoheemen, line 7:
```suggestion
Het onafhankelijke Ublad geeft een kleurrijk verslag van wat er speelde op de universiteit, de stad en het studentenleven door middel van artikelen, foto’s en cartoons. De afbeelding die is gebruikt voor OCR is voor elke pagina bijgevoegd zodat u altijd het originele bronmateriaal kunt raadplegen.
```

backend/corpora/ublad/images/ublad.jpg
JeltevanBoheemen, line 1: Leuke afbeelding, nice.

backend/corpora/ublad/ublad.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+from ianalyzer_readers.readers.html import HTMLReader
+from ianalyzer_readers.readers.core import Field
+from ianalyzer_readers.extract import html, Constant
+
+from bs4 import BeautifulSoup, Tag
+
+import locale
+locale.setlocale(locale.LC_ALL, 'nl_NL.UTF-8')
```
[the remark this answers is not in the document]
JeltevanBoheemen: I agree with Luka, this should not live in runtime code. My guess it is used in indexing?

JeltevanBoheemen (commented):
Few typos/language use suggestions. See Luka's review for a few code issues that should be resolved. Good work, nice that it's (almost) done!

backend/corpora/ublad/ublad.py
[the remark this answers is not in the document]
Meesch: It is; where would I put it then? Or what should I restore the locale to exactly?

backend/corpora/ublad/ublad.py
```diff
@@ -0,0 +1,253 @@
...
+    description = 'The print editions of the Utrecht University paper from 1969 until 2010.'
+    description_page = 'ublad.md'
+    min_date = datetime(year=1969, month=1, day=1)
+    max_date = datetime(year=2010, month=12, day=31)
+
+    data_directory = settings.UBLAD_DATA
+    es_index = getattr(settings, 'UBLAD_ES_INDEX', 'ublad')
+    image = 'ublad.jpg'
+    scan_image_type = getattr(settings, 'UBLAD_SCAN_IMAGE_TYPE', 'image/jpeg')
+    allow_image_download = getattr(settings, 'UBLAD_ALLOW_IMAGE_DOWNLOAD', True)
```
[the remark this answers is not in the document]
Meesch: `scan_image_type` is used in `GetMediaView`  to open the image type correctly, `allow_image_download` is used in the frontend to allow downlaods and the default is `false`. Or am I missing something?

Meesch:
I think I have addressed all your comments, see if you agree with my solutions. No worries if this does not make it into the next release!

backend/es/es_index.py
```diff
@@ -93,6 +95,15 @@ def populate(client, corpus_name, corpus_definition, start=None, end=None):
...
     corpus_server = settings.SERVERS[
         settings.CORPUS_SERVER_NAMES.get(corpus_name, 'default')]
+
+    # set corpus-specific locale
+    try:
+        corpus_locale = settings.CORPORA_LOCALES[corpus_name]
+        if corpus_locale:
+            locale.setlocale(locale.LC_ALL, corpus_locale)
+    except:
+        pass
```
lukavdplas, line 105: This is a catch-all `except` (i.e. it's not checking for specific error types); using `pass` as the body means that the script will not only move past the error but it will be completely undetectable. This is a big no-no, as it can create hidden bugs. In this case, if an unexpected error occurs during the indexing setup, you want to know what happened (and you _want_ the script to break so you can fix it first).

In most cases, a try/except block should look for *specific* errors that are expected under the circumstances. If it's a handler for any kind of error (usually because you're evaluating code that is not under the control of the current script, e.g. imported code, user code, etc.), the program should communicate what happened; usually by writing to the log or stderr.

In this case, it looks the block is meant to catch a `AttributeError` or `KeyError`, so a more specific `except` could be used. Both of those errors could also be avoided by providing appropriate fallbacks with `getattr` / `dict.get` during the lookup on line 101;  so you would not need an except clause at all.

backend/es/es_index.py
lukavdplas, line 1: I don't really agree with this solution. If locale switching is implemented in the application-wide indexing procedure, it increases the complexity of how indexing works generally; unless it's a common problem, it would be better to solve this problem in the corpus class, to keep the application more modular.

In this case, I expect it's possible to set and reset the locale in the `documents()` method of the corpus class instead. Something like this:

```python
def documents(self, sources=None):
    #set locale
    for doc in super().documents(sources):
        yield doc
    # reset locale
```

If it is *is* necessary to do this here, please make sure it's properly documented:

- This change adds a new setting, which should be documented in the [settings documentation](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/documentation/Django-project-settings.md)
- If this option is useful for other corpora, it should be included in the [documentation on writing Python corpora](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/documentation/Writing-a-corpus-definition-in-Python.md)
- There should be a unit test (or several) for the new functionality. As it is, it's quite likely that the functionality would be inadvertently broken or deleted at some point, since it will serve no apparent function; that is, no test will break if you remove it.

backend/corpora/ublad/ublad.py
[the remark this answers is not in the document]
lukavdplas: Yes, the values make sense, but you could just write the value directly, i.e.

```python
scan_image_type = 'image/jpeg'
allow_image_download = True
```

This definition allows the values to be configured in `settings`, but why? That would make sense if, for instance, image download would need to be disabled in some environments but not in others.

lukavdplas (changes requested):

lukavdplas (approved):

Issue #1605, opened 2024-06-13
Document highlights are CSS-only


lukavdplas:
Highlights in search results are only implemented as a CSS class, so they are not exposed to assistive technologies. The [`<mark>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) is designed for exactly this purpose and is more accessible.

To do:
- Update highlight queries so they specify the `<mark>` tag instead of a `<span class="highlight">`.
- Update the CSS rule that selects `.highlight` to select `mark`.

PR #1611, opened 2024-06-20
Feature/display entities

### LINKED ISSUE #1543: Endless `formatInnerHtml` loop
### What went wrong?

During debugging for #956 I put a breakpoint in `document-view.component.formatInnerHtml`, and noticed that the template keeps triggering the function, so that every field is processed endlessly. The call stack indicates that the `RouterStoreService` is responsible. We still seem to have a feedback loop.

Screenshot 2024-04-12 at 15 02 47

I expect that this function would be triggered *once* for each field.

### Screenshot

_No response_

- [ ] https://ianalyzer.hum.uu.nl
- [X] a local server

### Version

develop

### Steps to reproduce

Start up I-Analyzer, navigate to a corpus, put a breakpoint on line 108 of `document-view.component.ts`, and open a document view.

### LINKED ISSUE #1609: Highlight named entities in a document, or list named entities in the document preview

BeritJanssen:
Close #1609. Also close #1543 by implementing formatting of innerHTML through pipes. This branch implements fetching and displaying named entities from an elasticsearch index with an `annotated_text` field.

https://github.com/UUDigitalHumanitieslab/I-analyzer/assets/11174072/d7cba18c-4951-4d62-9348-fbd7e5ebab8a


frontend/src/app/document/entity-legend/entity-legend.component.html
```diff
@@ -0,0 +1,3 @@
+<div *ngFor="let entity of entities">
+    <span class="dot" [ngClass]="entity | formatEntityClass"></span><span>{{entity}}</span>
+</div>
```
lukavdplas, line 3: Can you think of a way to make the different types of entities readable to colorblind users?

BeritJanssen: I chose a colourblind-friendly palette. Do you think the colour differences would be too subtle for highlights?

BeritJanssen: I used a colorbrewer from adobe to come up with colours which might be suitable for highlights (the original colour palette was borrowed from the visualization, but for text background they don't work too well), and came up with this:
Screenshot 2024-06-20 at 17 33 23

lukavdplas: As far as I know, colorblind-friendly palettes are useful for the most common forms of colorblindness, but it's impossible to come up with a set of colors that works for all visually impaired users, which is why it's recommended that color is never the _only_ means to distinguish an element - see [WCAG 2.1 1.4.1](https://www.w3.org/TR/WCAG21/#use-of-color).

In this case, perhaps the component could also use icons to indicate the type of entity?

BeritJanssen: A solution I've come up with now: using flags at the start of the highlight to indicate the entity type:
Screenshot 2024-06-24 at 13 28 01

Alternatives I've considered:
- striping the highlights in different ways - this I thought would make the foreground text less visible
- using text balloons so users could hover to understand the type - this would make the interface less useable to color-blind users, as they cannot have the "at a glance" overview of which entities were found, and might also obstruct other users with pop-ups while reading the text
- using superscript icons after the highlight - I think superscripts would interrupt textflow more

backend/addcorpus/validation/creation.py
```diff
@@ -122,6 +122,13 @@ def validate_name_is_not_a_route_parameter(value):
             f'{value} cannot be used as a field name, because it is also a route parameter'
         )
 
+
+def validate_name_has_no_ner_suffix(value):
+    if value.endswith('_ner'):
+        raise ValidationError(
+            f'{value} cannot be used as a field name: the suffix `_ner` is reserved for annotated_text fields'
+        )
```
lukavdplas, line 130: Neat!

Since field names must be slugs, would it be possible to use a suffix that doesn't overlap? (e.g. `:ner` instead of `_ner`)

frontend/src/app/shared/pipes/geo-data.pipe.ts
lukavdplas, line 1: Nice to have this as a separate pipe!

frontend/src/app/shared/toggle/toggle.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="toggle-container">
+    <input type="checkbox" (click)="toggleButton()"/>
+    <label class="slider"></label>
+</div>
```
lukavdplas, line 4: This input is missing a an accessible label (or the label is missing text?)

frontend/src/app/document/entity-legend/entity-legend.component.html
lukavdplas, line 3: An issue with the legend specifically is that assistive technologies will parse it as a single chunk of text. The empty `<span>` elements will be ignored in the accessibility tree, so the inner structure will appear the same as:

```html
<div>
    Person Location Miscellaneous Organization
</div>
```

So this element is missing some meaningful information about its role or internal structure, including how each category relates to the content.

In this case, the element could use some semantic tags. For example, it could be marked up like this:

```html
<footer>
    <p>Legend</p>
    <ul>
        <li>Person</li>
        <li>Location</li>
        <li>Miscellaneous</li>
        <li>Organization</li>
    </ul>
</footer>
```

(or use [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) if you're feeling fancy).

That's still missing information about how "Person" actually relates to the content, though, but that depends on how the highlights are implemented.

backend/es/views.py
```diff
@@ -98,3 +100,84 @@ def _save_query_done(self, query, results):
...
+            input_text = self.substitute_annotation_with_tag(
+                annotation, input_text, entity_classes)
+        return input_text
+
+    def substitute_annotation_with_tag(self, annotation: tuple, input_text: str, entity_classes: set) -> str:
+        entity_class = annotation[1][1:-1]
+        annotated = annotation[0][1:-1]
+        entity_classes.update({entity_class})
+        return input_text.replace(
+            ''.join(annotation), f'<span class="entity-{entity_class.lower()}">{annotated}</span>')
```
lukavdplas, line 183: #1605 also applies here: the `<mark>` element is semantically appropriate here, and more accessible than `<span>`.

backend/addcorpus/validation/creation.py
BeritJanssen: `SlugField` has a built-in validator that only allows numbers, dashes and hyphens, doesn't it? I didn't think the `validators` argument when initializing a `models.SlugField` overrides this basic validator, but is added to it.

BeritJanssen: Nevermind, the validators do indeed replace it. When starting on the project, I remember using a colon as a separator, and running into the `SlugField` in-built validator. It's very likely that back then we didn't have the validator override in place yet.

lukavdplas: Oh, good to know!

I actually meant: as I understand, this validation is meant to prevent a situation where a corpus has two fields like `content_translated` and `content_translated_ner`, where the latter is just a normal content field. This is a problem because the named entity field for `content_translated` would also be called `content_translated_ner`.

You could keep the slug validation as-is, and then implement "special" suffixes with non-slug characters. So `content_translated` would have a named entity field `content_translated:ner`. So there is never any possible overlap between the slug name and any reserved suffixes.

frontend/src/app/document/document-popup/document-popup.component.html
```diff
@@ -1,9 +1,12 @@
 <ng-container *ngIf="document">
     <p-dialog [(visible)]="visible" width="100%"
-        [responsive]="true" [maximizable]="true" [dismissableMask]="true" [draggable]="true" [resizable]="false" [blockScroll]="true"
-        header="Document {{document.position}} of {{page.total}}">
+        [responsive]="true" [maximizable]="true" [dismissableMask]="true" [draggable]="true" [resizable]="false" [blockScroll]="true">
+        <ng-template pTemplate="header">
+            <span>Document {{document.position}} of {{page.total}}</span>
+            <span *ngIf="showNEROption" iaBalloon="Show named entities for this document">NER<ia-toggle (toggled)="toggleNER($event)"></ia-toggle></span>
```
lukavdplas, line 6: Two comments here:

- The label "NER" is not clear. The tooltip text is clearer, but I expect most users would still not know what NER stands for, because they're not familiar with the abbreviation. I would recommend against using this abbreviation anywhere in I-analyzer, since it's not common knowledge. In this case, "show named entities" would be a clearer label.  (NB: if the input has to use the shortest possible label due to a lack of space, then this may not be the right location for it.)
- Currently, NER can be toggled from here in the document popup, but there is no switch in the standalone page for a document. I don't think there is a reason why NER could not be available in that view?

frontend/src/app/shared/toggle/toggle.component.html
BeritJanssen: The `<label>` element is here purely to be able to attach pseudo-elements to it from css. Adding text will break the toggle button. The description of what the toggle button does should come from the component in which it is used; the "on / off" state is reflected by screen readers as it is built on top of an `<input>` element.

lukavdplas: Ah, I see. In that case:

- This element is not an actual label, so the semantics are misleading. If it's just for CSS, you can use a generic element instead (`<div>` or `<span>`). If that is not possible, it should set `role="generic"`.
- While the _value_ of the checkbox is machine-readable, the _label_ is not. That association exists visually, but will not show in the accessibility tree. For screen reader users, this creates the issue that it may be unclear what the checkbox is meant to do. Unlabelled form elements also make it harder to navigate a page using a screen reader (as you could otherwise use the overview of form elements).

The most typical technique for providing accessible labels is [using a `<label>` with "for" attribute](https://www.w3.org/WAI/WCAG22/Techniques/html/H44.html) but the component hierarchy may get in the way here. You could embed the input inside the `<label>` elements; see some examples in [MDN's documentation on `<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label). (I'm not sure if that also works when the input is not a direct child of the label, though.) This technique also has the advantage of increasing the clickable area.

An alternative technique would be to [use an `aria-labelled-by` attribute](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA16.html).

PR #1615, opened 2024-06-20
Remove IsAuthenticated permission from media views

JeltevanBoheemen:
Can all issues be this simple forever please?


BeritJanssen (approved):
Whoohoo!

PR #1618, opened 2024-06-28
Do not provide a default JMIG_DATA_URL

ar-jan:
Originally `JMIG_DATA` was a URL, which I changed to also accept a local path. Now it's either `JMIG_DATA_URL` or `JMIG_DATA`, so we can't have a default URL or `JMIG_DATA` is never used.


ar-jan:
Actually now that data_directory is part of the corpus model it has to be a directory so we cannot (ab)use it for the data source.

BeritJanssen (approved):
This might work, but please check that it's possible to run with `data_directory` being `None` before merging.

PR #1619, opened 2024-07-04
Feature/cache actions

BeritJanssen:
Close https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1489: use GitHub registry to cache images.


.github/workflows/backend-test.yml
```diff
@@ -13,15 +13,45 @@ on:
...
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
```
JeltevanBoheemen, line 33: I understand this is the user triggering the workflow. Does this need any setup for the developers?

JeltevanBoheemen (approved):

.github/workflows/backend-test.yml
BeritJanssen: Nope, this is basically a user pre-configured by the action which has rights (for this repo) to write to this registry. We would need to setup username / passwords in GitHub secrets if we used a non-GitHub registry, I believe.

PR #1621, opened 2024-07-04
use params.get instead of lodash get

### LINKED ISSUE #1297: Nearest neighbours: neighbours input resets on confirm
### What went wrong?

When you update the number of nearest neighbours, the graph is updated correctly, but the input field resets to the default value.

### Expected behavior

The input field should stay at the current value.

### Screenshots

[Screen capture. Shows the user updating the input field. On confirmation, the graph updates correctly but the input field resets](https://github.com/UUDigitalHumanitieslab/I-analyzer/assets/43678097/225dde85-fb4b-4c70-8e82-6e6ffb43e0ab)

### Which version?

https://ianalyzer.hum.uu.nl

### To reproduce

- Go to https://ianalyzer.hum.uu.nl/word-models/parliament-netherlands?query=test&neighbours=5
- Update the number of nearest neighbours to include
- Press "confirm"

BeritJanssen:
Close #1297.


JeltevanBoheemen:
Good to see this fixes it. Any idea _why_? :)

BeritJanssen:
Yes: params is not a simple object, but a `Params` object, which has a method `.get`. Otherwise, we'd need to do `_.get(params.params, ...)`, but since we already have a default for `this.neighbours`, this is not necessary.

lukavdplas:
> Yes: params is not a simple object, but a Params object, which has a method .get. Otherwise, we'd need to do _.get(params.params, ...), but since we already have a default for this.neighbours, this is not necessary.

Did you test this? This description sounds like [ParamMap](https://v17.angular.io/api/router/ParamMap), not [Params](https://v17.angular.io/api/router/Params).

BeritJanssen:
Yes, indeed, the object that gets passed in is a ParamMap (in ParamDirective: `this.route.queryParamMap.subscribe...`). I went ahead and corrected the typing.

lukavdplas (approved):

Issue #1623, opened 2024-07-04
Query in download CSV not rendered


lukavdplas:
If you download search results for a query that included a search term, the CSV is supposed to have a column `query` which contains the search term. Right now, it will instead show a string representation of the `query` Python module in the backend.

PR #1625, opened 2024-07-04
Query in download CSV not rendered #1623 

supercoder-dev:
Fix query text rendering in CSV downloads by updating write_output to include the query in headers and data, and modifying set_long_wide_format to handle queries in wide format, resolving issue #1623.


oktaal:
Adding a nonsensical `"http://example.com/schema/corpus.json"` schema url and some whitespace bullshit. Such code. Much wow.

PR #1632, opened 2024-07-18
resize times image

lukavdplas:
Very minor point: it annoyed me that the image for the Times corpus had such a large file size for a thumbnail (also compared to other corpus images) and had a colour grade that did not match the page. Resized it from ~1,5 MB to ~150 KB and changed the colours.


JeltevanBoheemen (approved):
This does not really warrant a review request, but okay: approved :)

Issue #1633, opened 2024-07-18
"query in context" field does not work


lukavdplas:
### What went wrong?

If you try to download a search results with snippets ("query in context"), the download will fail.

### What did you expect to happen?

The download should not raise an error.

When this feature still worked, it would add a variable number of untitled columns as the last columns in the CSV. Those would contain snippets of text surrounding matches to the query.

### Where did you find the bug?

- [X] https://ianalyzer.hum.uu.nl
- [ ] https://peopleandparliament.hum.uu.nl
- [ ] https://peace.sites.uu.nl
- [ ] a server hosted elsewhere (i.e. not by the research software lab)
- [X] a local server

### Version

5.9.0

### Steps to reproduce

- Go to https://ianalyzer.hum.uu.nl/search/troonredes?query=nederland&highlight=200 - or go to a search page for a corpus, enter a query term and turn on highlighting
- In the field selection for the download menu, click "select all fields" and then select "query in context"
- Click "download"

lukavdplas:
The error is in `backend/download/create_csv.py`. This module uses `csv.DictWriter` to create CSVs, but this writer does not support variable column numbers.

The DictWriter must receive a list of all fieldnames up front. We don't have that list, because each snippet of a document gets a new column; that means we don't know the number of query-in-context columns before we iterate through the data.

I think this may have worked in the past because all the CSV data was loaded in memory before writing the file. I remember that at some point, we rewrote the download module to avoid this, so it won't overload the memory for large files.

Some solutions, none of them very attractive:

- Use `csv.writer` instead of `csv.Dictwriter`, which won't complain about a variable number of columns, or about some columns lacking headers. However, the resulting CSV is not great. Some readers will parse them without issue, some won't.
- Add headers for an abundant number of context columns (e.g. 100). Cap off the number of snippets at this number.
- Reformat the CSV to have a single `context` column, which contains all snippets separated by newlines.

lukavdplas:
By the way, I was wondering why this isn't picked up by unit tests. The reason is that this test fixture is misleading:

https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/a0da9e1ea0a5959602fdc6a5adf96df1755f8ba4/backend/download/tests/test_csv_results.py#L44-L49

The name suggests that it's creating a CSV with highlight snippets, but that doesn't match the parameters passed to the `search_results_csv` function. So it appears like we're testing CSVs with content snippets, but that's not actually happening.

PR #1635, opened 2024-07-19
Feature/download tab

### LINKED ISSUE #674: Bigger search bar?
A suggestion:

image

For reference, this is the current layout:

image

### LINKED ISSUE #1029: Download results button occasionally disabled
If you go to the visualisation tab and update the query or filters, the download button will be disabled.

It used to be that updating the query would always take you to the results tab as well, so this is a recent phenomenon. This happens because the frontend has not requested results, and thus the `resultCount` is 0.

This is a minor issue, but it's not great UI to disable a button when there is no reason why it cannot be executed. And while users can still download by switching to the results tab, they have no way of knowing that.

Could be resolved by:
- Making a search request (with `size = 0` for speed) to probe the total number of results 
- Implementing the layout change I propose in #674, to move the download button into the results tab.

### LINKED ISSUE #1118: Add field selection form to download options dialog
> Suggestion: add the field selection form to the dialog that pops up when you select `Download CSV`. I think it is confusing that there are two separate menus to configure the download.

Hm, the problem is that, the way everything is set up right now, field selection has to be done _before_ you fetch the results, while the popup window shows options for formatting _after_ results have been fetched. This distinction is more apparent for long downloads: you choose the field when requesting the download, and the encoding when you download the result.

It is indeed not very intuitive to users. We could resolve it by always fetching all fields initially, so the user can make a selection for the final formatting? That is probably more convenient for users, though it does increase the storage size of the downloads on the server.

_Originally posted by @lukavdplas in https://github.com/UUDigitalHumanitieslab/I-analyzer-django/issues/36#issuecomment-1551364888_

### LINKED ISSUE #1310: Add download tab to search page

- https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1118 : downloading search results splits the user flow into picking the search fields and picking other parameters. This is not intuitive: all of this should be handled in a single form.
- https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/674 : the location of the download button leaves less space for the query input. It also results in a weird flow. You download a representation of the results - typically after looking at them, but its location suggests that this control is part of the query setup. However, moving it down to the search results page clutters the page.

Add a third tab next to "Results" and "Visualisations", with "Download", and move the download form / button there. Quick mock-up:

screenshot showing a "download" tab with a form for fields, highlighting and encoding

- #674 suggests an alternative design to move the download button into the search results tab instead
- my suggestion in #1118 was move the field selection into the dialogue window.

lukavdplas:
This moves the download menu into a separate tab.

The download tab includes a form with all options for the download; it replaces both the "download" button on the top (with the attached dropdown menu) and the options dialogue window. (The options dialogue is still used in the download history, though.)

close #1310
close #1029
close #674
close #1118 (this suggested an alternative approach)

Screen capture:

[screencast of I-analyzer. Shows a "download" tab in the search results for a corpus. This contains a form, with a message that there are many results and the user will receive an email when the download is complete. The user then types in a query to limit the number of results, and the form is updated for a direct download.](https://github.com/user-attachments/assets/16085c18-7f0a-4611-ba01-4a25ace1a6c6)

Notes:
- The download form includes the option to select "query in context". This works on the frontend side, but the functionality is broken on the backend (#1633). I've hidden the option in the form for now.
- I've included the test I used to isolate #1633 and marked it xfail.
- This changes the way tabs are encoded in the URL, with an explicit `tab` parameter. Old routes will continue to work.
- Also added `tags` and `context` as forbidden field names. This isn't because of any changes in this PR, I just noticed they should have been added.


frontend/src/app/download/download.component.ts
```diff
@@ -40,66 +55,61 @@ export class DownloadComponent implements OnChanges {
...
-                mappingType: null,
-            } as unknown as CorpusField);
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.corpus) {
+            this.availableCsvFields = _.filter(this.corpus?.fields, 'downloadable');
+            this.selectedCsvFields = _.filter(this.corpus?.fields, 'csvCore');
+        }
+        if (changes.queryModel) {
+            this.totalResults?.complete();
+            this.resultsConfig?.complete();
```
JeltevanBoheemen, line 74: Careful closing of the observables here, I wouldn't have thought to do so if replacing them with a new stream.
Out of curiosity, did problems arise when not closed?

frontend/src/app/download/download.component.ts
```diff
@@ -40,66 +55,61 @@ export class DownloadComponent implements OnChanges {
...
-     */
-    public chooseDownloadMethod() {
-        if (
-            this.resultOverview.resultsCount < this.directDownloadLimit ||
-            this.downloadLimit === undefined
-        ) {
-            this.directDownload();
+    onHighlightChange(event): void {
+        if (event.target.checked) {
+            this.resultsConfig.setParams({ highlight: 200 });
```
JeltevanBoheemen, line 92: Magic number

frontend/src/app/download/download.component.ts
```diff
@@ -142,11 +146,40 @@ export class DownloadComponent implements OnChanges {
             });
     }
 
-    private getCsvFields(): CorpusField[] {
+    private enableDirectDownload(totalResults: number) {
```
JeltevanBoheemen, line 149: Missing a return type, though obvious from the code. Could also omit the parameter and use `this.totalResults`?

frontend/src/app/search/results-sort/search-sorting.component.spec.ts
```diff
@@ -1,13 +1,13 @@
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
-import { mockCorpus } from '../../mock-data/corpus';
-import { commonTestBed } from '../common-test-bed';
-import { QueryModel } from '../models';
+import { mockCorpus } from '../../../mock-data/corpus';
+import { commonTestBed } from '../../common-test-bed';
+import { QueryModel } from '../../models';
```
JeltevanBoheemen, line 4: The imports for models and services would benefit from path mapping in `tsconfig.json`

frontend/src/app/search/search-tabs.ts
```diff
@@ -0,0 +1,37 @@
...
+            return { tab: 'search-results' }
+        };
+    }
+
+    protected stateToStore(state: SearchTabState): Params {
+        if (state.tab === 'search-results') {
+            return { tab: null };
+        } else {
+            return state;
+        }
```
JeltevanBoheemen, line 35: Familiarizing myself with the `store` stuff, which does look very neat. 
I wonder if something like a 'default state' would be beneficial to add as an option in `StoreSync`? In this case if the tab is `search-results`.

JeltevanBoheemen (approved):
As always: neat code, good tests.
The download component is a bit heavy, but the functionality all directly concerns downloading, so that's fine.

The new look for the download tab is a lot better.

frontend/src/app/download/download.component.ts
lukavdplas: No, this is more "good practice", because we're creating observables in `ngOnChanges`, it's good to close them when new input comes in.

But this is triggered when the input receives a new `queryModel`. That's really an edge case. (See https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1247#issuecomment-1932377736 for more detail.)

frontend/src/app/search/search-tabs.ts
lukavdplas: Hm, good point. Filters have something like that, but perhaps it could be generalised to more models.

frontend/src/app/download/download.component.ts
lukavdplas: 🧙‍♂️

PR #1636, opened 2024-07-24
Feature/version commands

JeltevanBoheemen:
Provides commands for bumping version numbers and updating CITATION.cff alongside it.
This _could_ be run in a GitHub action, but I prefer doing this manually, without the need to make a separate commit/push. Thoughts on this @lukavdplas @BeritJanssen ?


update_citation.py
```diff
@@ -0,0 +1,38 @@
...
+VERSION = None
+TODAY = datetime.today().strftime('%Y-%m-%d')
+
+with open(PACKAGE_FILE, 'r') as package_file:
+    package_json = json.load(package_file)
+    try:
+        VERSION = package_json.get('version')
+    except KeyError:
+        print('No version found in root package.json')
+        VERSION = '0.0.0'
```
lukavdplas, line 23: Setting the version in the citation to 0.0.0 in this case doesn't look like something we would actually want. I would expect one of the following:

- raise an error
- remove the `version` key from CITATION.cff

JeltevanBoheemen: Yeah you're right. I think just raise an error to be honest. This is not a generalised solution; not finding a version is not supposed to happen.

BeritJanssen:
Nice! I have to say that I don't clearly have an opinion either way. One thing we have to do anyway by hand is create the release branch. Arguably, you could save one step of work with triggering `yarn version` through a GitHub action. But then to new developers it might be obscure that something like this is actually happening upon a push to a `release/*` branch. So perhaps just adding the command and documenting it would be sufficient. In which case, the `release.yaml` action could be deleted.

Is it worthwhile adding this script to the cookiecutter? (After just having a discussion about the command-heaviness of the cookiecutter, this doesn't seem the optimal solution, but I do think this script is interesting for other repositories, too.)

jgonggrijp:
Regarding automation: there is also the option of a commit hook. No need to push to GitHub first, but also no need to remember to run a command.

As for the command-heaviness of the cookiecutter, I agree. I think dockerizing the cookiecutter wholesale would help a lot on this front. I think you said you wanted to have another go at that, @BeritJanssen? Otherwise, I might give it a try.

PR #1637, opened 2024-07-25
Feature/entity documentation

### LINKED ISSUE #1630: Add documentation on named entities
Support for named entity annotations is currently not documented at all. At minimum, the following documentation should be added.

In the developer documentation:
- While you could discover that I-analyzer supposed named entity annotations by reading the source code, the documentation should mention this up front.
- The documentation should be clear on how named entity annotations can be added to a corpus. Based on https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/1611 , I understand that this is either not implemented yet or intended to be handled through an external service; either of those options should be clearly documented.
- If the intention is that named entity annotations are made by an external service, the documentation must provide precise expectations of what this service is meant to do to provide compatible annotations.
- Usage of the annotation plugin in Elasticsearch must be documented in the installation instructions.

In the user manual:
- The manual should provide a brief explanation of what a named entity is and link to more information.
- The manual should give an explanation of the categories of named entities that are implemented in I-analyzer, with examples for each.
- The manual should describe what functionality is supported for named entities.

BeritJanssen:
Close #1630. Factored out the entity toggle, trying out a new Angular feature: output signals. Also adjusted the paragraph pipe, to fix this problem:
Screenshot 2024-07-25 at 12 01 37

After the change:
Screenshot 2024-07-25 at 12 02 52


frontend/src/app/document/entity-toggle/entity-toggle.component.html
```diff
@@ -0,0 +1,6 @@
+<button class="button" type="button" (click)="showNamedEntityDocumentation()">
+    <span class="icon">
+        <fa-icon [icon]="actionIcons.helpAlt" aria-label="help"></fa-icon>
+    </span>
+</button>
+<em iaBalloon="Show named entities for this document">Show named entities<ia-toggle (toggled)="toggleNER.emit($event)"></ia-toggle></em>
```
lukavdplas, line 6: Two comments:

- The tooltip seems unnecessary here, it doesn't actually add any information.
- This component is providing the label for the toggle element but that link isn't encoded in a machine-readable way. I suggested some possible implementations for that in [this comment on your earlier PR](https://github.com/UUDigitalHumanitieslab/I-analyzer/pull/1611#discussion_r1662380657).

frontend/src/app/shared/pipes/paragraph.pipe.ts
```diff
@@ -1,22 +1,21 @@
...
     }
 
-    addParagraphTags(content: string | string[]) {
+    addParagraphBreaks(content: string | string[]) {
         const paragraphs = typeof content === 'string' ? content.split('\n') : content;
         if (!paragraphs || paragraphs.length === 1) {
             return content;
         }
-        return paragraphs.map(p => `<p>${p}</p>`).join(' ');
+        return paragraphs.filter(p => p !== '').join('<br><br>');
```
lukavdplas, line 18: Paragraphs `<p>` and linebreaks `<br>` are different things; linebreaks are used for anything that isn't a paragraph break, e.g. in postal addresses or poems. This concerns actual paragraphs in a text, though, so it's not appropriate to use `<br>` here.

Using paragraphs appropriately makes the text much more user-friendly for screen readers, especially if you're close-reading a longer document. See also [accessibility concerns for `<br>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility).

frontend/src/assets/manual/en-GB/namedentities.md
```diff
@@ -0,0 +1,10 @@
+In some corpora, you have the option to view named entities. If this option is available, a toggle button labeled *Show named entities* is visible in the document overview, above the document content and metadata.
+
+Activating the toggle will show entities of four types:
+- Persons
+- Locations
+- Organizations
+- Miscellaneous
+
+### Disclaimer
+Note that named entities have been extracted through machine learning using [the TextMiNER library](https://github.com/CentreForDigitalHumanities/TextMiNER). As such, the annotated entities should not be taken at face value, and only be used for exploration purposes.
```
lukavdplas, line 10: Good call to have this disclaimer!

Some comments:
- I think it would be good to be a bit more clear about why machine learning means you shouldn't take classifications at face value.
- Technically, nothing in I-analyzer enforces that you use the textminer library, so I wouldn't hard-code that into the manual here. It would make more sense if the library was linked as part of the corpus documentation, like we do for word models.
- Methodologically, I don't agree that the output of machine learning models should only be used for exploration. Automated classifiers are used all the time in research, and it's possible to use their output data responsibly.

```suggestion
### Disclaimer
Named entities are automatically assigned using machine learning. As such, the annotations can include mistakes, and should not be taken at face value.

If a corpus has named entity annotations, the info page for the corpus will provide more information on how the models were trained. We recommend being critical about the accuracy and possible biases of this process if you intend to use named entities in your research.
```

frontend/src/assets/manual/en-GB/namedentities.md
```diff
@@ -0,0 +1,10 @@
+In some corpora, you have the option to view named entities. If this option is available, a toggle button labeled *Show named entities* is visible in the document overview, above the document content and metadata.
+
+Activating the toggle will show entities of four types:
+- Persons
+- Locations
+- Organizations
+- Miscellaneous
+
+### Disclaimer
```
lukavdplas, line 9: This should be a second-level header (a third-level header would be appropriate within a section that has a second-level header).

```suggestion
## Disclaimer
```

lukavdplas (changes requested):
Neat! :)

The refactor looks good, explanations are clear. I've added some suggestions for small improvements; feel free to merge if you've handled them.

BeritJanssen:
I managed to solve the issue to display icons without breaking the paragraph: two pipes in a row, first a pipe which adds the markup for the entities, then the paragraph pipe. The `icon()` function turned out to do the trick for the entity pipe. Rather than inserting `fa-icon` elements, it inserts `svg` elements. I just wonder whether these icons will throw off screen readers now. 🤔 I didn't see any way to add classes or attributes to the resulting `svg` element, but the `mark` tags will have labels. The toggle button now also has an accessible label.

Issue #1573, opened 2024-05-16
Upload corpus image + documentation


lukavdplas:
The backend API supports adding a corpus definition, but it still needs support for:

- [ ] changing the image for a corpus
- [x] adding / changing documentation pages

These are not directly essential for creating a working corpus, but they are the kind of thing that we would expect a curator (e.g. a researcher) to manage.

PR #1640, opened 2024-08-07
Edit corpus documention in API

lukavdplas:
This update changes the backend API to allow uploading corpus documentation (part of https://github.com/UUDigitalHumanitieslab/I-analyzer/issues/1573). It also makes some quality improvements to permissions in corpus-related views.

The corpus documentation view now has create/update/delete endpoints (accessible to admin users). Some changes to the view to make that more convenient:
- The corpus name is now an optional query parameter, e.g. `/api/corpus/documentation/dbnl/` -> `/api/corpus/documentation/?corpus=dbnl`. It's also included in the JSON data, so you can create/update on `/api/corpus/documentation/`.
- The view includes both the original template for the documentation, and the rendered version (the latter is read-only).
- The order of pages is now included as a (read-only) `index` field in the JSON, rather than implicit in the array order.
- If the corpus has word models documentation but no models, they are filtered out in the frontend instead of the backend.
- Admin users can now view the documentation of inactive corpora

I added two new permissions for corpus-related views: `IsCurator` and `IsCuratorOrReadOnly`. Also, the name `CorpusAccessPermission` was getting confusing (access to do what?), so I renamed that to `CanSearchCorpus`.

### DIFF
--- a/backend/addcorpus/serializers.py
+++ b/backend/addcorpus/serializers.py
@@ -121,13 +121,29 @@ class DocumentationTemplateField(serializers.CharField):
         return render_documentation_context(content)
 
 
+class CorpusConfigurationField(serializers.SlugRelatedField):
+    '''
+    Serialise a related corpus configuration as the corpus name.
+    '''
+
+    def __init__(self, **kwargs):
+        super().__init__('corpus__name', **kwargs)
+
+    def get_queryset(self):
+        return CorpusConfiguration.objects.all()
+
+    def to_representation(self, value: CorpusConfiguration) -> str:
+        return value.corpus.name
+
+
 class CorpusDocumentationPageSerializer(serializers.ModelSerializer):
     type = PrettyChoiceField(choices = CorpusDocumentationPage.PageType.choices)
     content = DocumentationTemplateField()
+    corpus = CorpusConfigurationField(source='corpus_configuration')
 
     class Meta:
         model = CorpusDocumentationPage
-        fields = ['corpus_configuration', 'type', 'content']
+        fields = ['id', 'corpus', 'type', 'content']
 
 
 class JSONDefinitionField(serializers.Field):
--- a/backend/addcorpus/serializers.py
+++ b/backend/addcorpus/serializers.py
@@ -65,6 +65,16 @@ class PrettyChoiceField(serializers.ChoiceField):
         key = super().to_representation(value)
         return self.choices[key]
 
+    def to_internal_value(self, data):
+        # If the data provides a display name, get the corresponding key.
+        # The browsable API sends keys instead of labels; use the original data if no
+        # matching label is found.
+        value = next(
+            (key for (key, label) in self.choices.items() if label == data),
+            data
+        )
+        return super().to_internal_value(value)
+
 class CorpusConfigurationSerializer(serializers.ModelSerializer):
     fields = FieldSerializer(many=True, read_only=True)
     languages = serializers.ListField(child=LanguageField())
--- a/backend/addcorpus/serializers.py
+++ b/backend/addcorpus/serializers.py
@@ -136,14 +136,32 @@ class CorpusConfigurationField(serializers.SlugRelatedField):
         return value.corpus.name
 
 
+class ChoiceOrderSerializer(serializers.ReadOnlyField, serializers.ChoiceField):
+    def __init__(self, choices, **kwargs):
+        super().__init__(choices=choices, **kwargs)
+        self.index = {
+            value: i
+            for i, (value, label) in enumerate(choices)
+        }
+
+    def to_representation(self, value):
+        key = super().to_representation(value)
+        return self.index[key]
+
+
 class CorpusDocumentationPageSerializer(serializers.ModelSerializer):
     type = PrettyChoiceField(choices = CorpusDocumentationPage.PageType.choices)
+    index = ChoiceOrderSerializer(
+        read_only=True,
+        source='type',
+        choices=CorpusDocumentationPage.PageType.choices,
+    )
     content = DocumentationTemplateField()
     corpus = CorpusConfigurationField(source='corpus_configuration')
 
     class Meta:
         model = CorpusDocumentationPage
-        fields = ['id', 'corpus', 'type', 'content']
+        fields = ['id', 'corpus', 'type', 'content', 'index']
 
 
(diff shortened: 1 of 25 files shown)


backend/addcorpus/serializers.py
```diff
@@ -121,13 +121,51 @@ def to_representation(self, value):
...
+    '''
+
+    def __init__(self, **kwargs):
+        super().__init__('corpus__name', **kwargs)
+
+    def get_queryset(self):
+        return CorpusConfiguration.objects.all()
+
+    def to_representation(self, value: CorpusConfiguration) -> str:
+        return value.corpus.name
```
JeltevanBoheemen, line 136: This looks overengineered to me, could it be solved with a [SlugRelatedField](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield)?

backend/addcorpus/serializers.py
```diff
@@ -121,13 +121,51 @@ def to_representation(self, value):
...
+    def __init__(self, choices, **kwargs):
+        super().__init__(choices=choices, **kwargs)
+        self.index = {value: i for i, (value, label) in enumerate(choices)}
+
+    def to_internal_value(self, data):
+        return self.choices[data]
+
+    def to_representation(self, value):
+        key = super().to_representation(value)
+        return self.index[key]
```
JeltevanBoheemen, line 152: Also slightly overengineered, I would suggest just using a calculated property on the model and serializing that, **BUT**: the reverse operation is nice, being able to send the index.

backend/addcorpus/tests/test_corpus_views.py
```diff
@@ -16,23 +16,43 @@ def test_no_corpora(db, settings, admin_client):
...
+    )
+    page_types = [page['type'] for page in sorted_and_filtered]
     assert page_types == ['General information', 'Citation', 'License']
 
     # should contain citation guidelines
-    citation_page = next(page for page in pages if page['type'] == 'Citation')
+    citation_page = next(
+        page for page in response.data if
+        page['type'] == 'Citation' and page['corpus'] == basic_mock_corpus
+    )
```
JeltevanBoheemen, line 36: Would `{'type': 'Citation', 'corpus': basic_mock_corpus} in response.data` suffice? For readability

backend/addcorpus/tests/test_corpus_views.py
```diff
@@ -16,23 +16,43 @@ def test_no_corpora(db, settings, admin_client):
...
 
     # check that the page template is rendered with context
     content = citation_page['content']
     assert '{{ frontend_url }}' not in content
     assert settings.BASE_URL in content
 
+def test_corpus_documentation_filter_list_view(admin_client, basic_mock_corpus):
+    response = admin_client.get(f'/api/corpus/documentation/?corpus={basic_mock_corpus}')
+    assert response.status_code == 200
```
JeltevanBoheemen, line 45: A very minor pet peeve, I prefer the `HTTP_200_OK` etc constants from drf, because I always forget which status is which.

JeltevanBoheemen: Don't go changing that by the way, the application is filled with just numbers as status code checks.

backend/addcorpus/views.py
```diff
@@ -1,78 +1,58 @@
...
-    permission_classes = (IsAuthenticatedOrReadOnly,)
+    serializer_class = CorpusSerializer
 
-    def get(self, request, *args, **kwargs):
-        corpora = Corpus.objects.filter(active=True)
-        filtered_corpora = filter_user_corpora(corpora, request.user)
-        serializer = CorpusSerializer(filtered_corpora, many=True)
-        return Response(serializer.data)
+    def get_queryset(self):
+        return self.request.user.searchable_corpora()
```
JeltevanBoheemen, line 23: Noice

backend/users/models.py
```diff
@@ -1,23 +1,39 @@
 import django.contrib.auth.models as django_auth_models
 from django.db import models
+from addcorpus.models import Corpus
 
 DEFAULT_DOWNLOAD_LIMIT = 10000
-
+PUBLIC_GROUP_NAME = 'basic'
```
JeltevanBoheemen, line 6: Should probably just change this to `'public'` at some point :)

frontend/src/app/corpus-info/corpus-info.component.ts
```diff
@@ -41,4 +46,12 @@ export class CorpusInfoComponent implements OnInit {
         return marked.parse(content);
     }
 
+    private sortPages(pages: CorpusDocumentationPage[]): CorpusDocumentationPage[] {
+        return _.sortBy(pages, 'index');
+    }
```
JeltevanBoheemen, line 51: This does not warrant a function, unless expected to do more in the future. 
The readable name is nice, but `sortBy` is intuitive enough that I prefer not bloating the component.

frontend/src/app/corpus-info/corpus-info.component.ts
```diff
@@ -41,4 +46,12 @@ export class CorpusInfoComponent implements OnInit {
...
+
+    private includePage(corpus: Corpus, page: CorpusDocumentationPage): boolean {
+        return page.type !== 'Word models' || corpus.wordModelsPresent;
```
JeltevanBoheemen, line 54: Add some parentheses to this expression to ensure to everyone that it does what is expected.

frontend/src/app/services/api.service.ts
```diff
@@ -223,14 +223,17 @@ export class ApiService {
     }
 
     // Corpus
-    public corpusDocumentation(corpusName: string): Observable<CorpusDocumentationPage[]> {
-        const url = this.apiRoute(
-            this.corpusApiUrl,
-            `documentation/${corpusName}/`
-        );
+    public corpusDocumentationPages(corpus?: Corpus): Observable<CorpusDocumentationPage[]> {
+        const queryParams = corpus ? `?corpus=${corpus.name}` : ''
```
JeltevanBoheemen, line 227: I suggest using [HttpParams](https://angular.dev/api/common/http/HttpParams)

backend/addcorpus/serializers.py
JeltevanBoheemen: s in, directly use the SlugRelatedField in the serializer instead of extending it.

PR #1643, opened 2024-08-13
Feature/frontend import shortcuts

JeltevanBoheemen:
These paths made sense to me. The imports were performed with a regex, so in some cases they don't really 'help', but still prefer having the consistent import style.

Only `tsconfig.json` is relevant for review. `compilerOptions` -> `baseUrl`, `outDir`, and `paths` were changed (to help you read through the autoformatter)


lukavdplas (approved):
:+1:

PR #1644, opened 2024-08-15
Feature/minor accessibility improvements

### LINKED ISSUE #1603: Tooltip messes up document table structure
In the `DocumentPreviewComponent`, the  `iaTooltip` is used on a `<th>` (table header) element. This seems to mess up the table structure (probably because of the pseudo-element created by the tooltip), which wasn't parsed properly when I tested it.

This can be fixed by applying the tooltip to a child of the header element. This fixed it for me:

```html
<th>
	<span iaTooltip="...">...</span>
</th>
```

### LINKED ISSUE #1605: Document highlights are CSS-only

lukavdplas:
This includes a few small changes for accessibility.

- Fixes the contentinfo landmark being used twice (by the element containing the page footer as well as within the footer). This landmark should be unique.
- Fixes tooltips interfering with the table structure for document data (close #1603)
- Use `<mark>` instead of `<span>` for highlights. (close #1605)
- Use appropriate header level for the filters section


frontend/src/app/footer/footer.component.html
```diff
@@ -1,4 +1,4 @@
-<footer class="footer">
```
JeltevanBoheemen, line 1: Hmm my intuition is that semantic html tags actually help accesibility. But this is not the case?

JeltevanBoheemen (approved):

frontend/src/app/footer/footer.component.html
lukavdplas: In general yes. In most sites, the `<body>` is layed out roughly like this:

```html
<header></header>
<main></main>
<footer></footer>
```

This is preferable to making three `<div>` elements, because this also assigns the semantic role for those elements; respectively `banner`, `main`, and `contentinfo`. Those roles also act as landmarks, so they can help users navigate the page. The issue was that I-analyzer was layed out like this:

```html
<ia-menu role="header"></ia-menu>
<main></main>
<ia-footer role="contentinfo">
	<footer></footer>
</ia-footer>
```

So we had a footer within a footer.

PR #1654, opened 2024-09-04
Jewishmigration: request data with API key

BeritJanssen:
This PR adds a couple of lines to send a get request with an authorization header in case an api key is defined in the settings. It also fixes a problem with `JewishMigration`'s `es_alias` being set to "peaceportal" (inherited from `PeacePortal` parent class).


ar-jan (approved):
Looks good; `HTTPBasicAuth` import can be removed again.

PR #1657, opened 2024-09-06
Use MapDataResults class to manage fetching, queryModel subscription

ar-jan:
Use MapDataResults class to manage fetching, queryModel subscription (based on WordcloudComponent and FrequentWordsResults).


frontend/src/app/models/map-data.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    fetch(): Observable<MapData> {
+        const field = this.state$.value.field;
+        if (!field) {
+            return of({ geoDocuments: [], mapCenter: null });
+        }
+
+        const getGeoCentroid$ = this.mapCenter
+            ? of(this.mapCenter)
```
BeritJanssen, line 43: Better to just name this variable `geoCentroid$`, in my opinion. I expect a getter function (i.e., callable) when a variable starts with `get`.

You could also give mapCenter the type of an `Observable<GeoLocation`> and then do:
```
const mapCenter$ = this.visualizationService.getGeoCentroid(field.name, this.query.corpus).pipe(share);
```
This uses the [share operator](https://rxjs.dev/api/index/function/share) which will basically do the same as you wrote here, i.e., only fetch the centroid if it isn't defined yet.

ar-jan: TIL, changed.

lukavdplas: Your previous solution was correct, this one is not.

The `share` operator is meant to handle situations like where you create an observable once but subscribe to it multiple times; like this:

```ts
const mapCenter$: Observable<any> = makeRequest();
mapCenter$.subscribe(doSomething);
mapCenter$.subscribe(doSomethingElse);
```

Using `subscribe` on an observable is akin to calling a function; the body of the observable is executed each time you subscribe to it. So you would be making the request twice here.

However, because you are assigning `mapCenter$` in the body of `fetch()`, this situation is more like this:

```ts
let mapCenter$: Observable<any> = makeRequest(); // or makeRequest().pipe(share())
mapCenter$.subscribe(doSomething);

mapCenter$ = makeRequest(); // or makeRequest().pipe(share())
mapCenter$.subscribe(doSomething);
```

The issue is not multiple subscribers, but the fact that we re-define `mapCenter$` every time `fetch` is called (i.e. every time the input is updated). 

By the way, the actual code is more like this:

```ts
function fetch(): Observable<any> {
	mapCenter$ = makeRequest();
	return mapCenter$;
}
// call fetch() whenever the query or the parameters update
result$ = combineLatest([queryModel.update, state$]).pipe(
	switchMap(() => fetch()),
	shareReplay(1),
).subscribe(doSomething());
```

So `result$`  [is already using a `shareReplay` operator](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/04a27ffe65f7015b70f2af860864d9969c9e0973/frontend/src/app/models/results.ts#L56-L66) on the output of `fetch`.

(For HTTP requests, you usually want to use `shareReplay` instead of `share`; this allows new observers to get the result if they subscribe after it arrived.)

Anyway, I think your original solution was sound; store `geoCentroid$` as a member of the class, and return the existing value if it's already defined. Keep in mind that the result does depend on the visualised field, however.

If you prefer a more declarative solution, I think something like this should work:

```ts
class MapDataResult {
	constructor() {
		// ...
		// update mapCenter$ based on state.visualizedField, ignoring query updates
		this.mapCenter$ = this.state$.pipe(
			switchMap(field => this.visualizationService.getGeoCentroid(field, this.query.corpus)),
		);
	}

	fetch() {
		const documents$ = this.visualizationService.getGeoData(...);
		return documents$.pipe(
			withLatestFrom(this.mapCenter$)
		);
	}
}
```

ar-jan: Thanks for the thorough explanation! I picked your last approach.

ar-jan:
Pushed an update using @lukavdplas suggestion.

PR #1658, opened 2024-09-11
Add try/except clause around streaming_bulk operation

BeritJanssen:
When indexing the JewishMigration corpus, the process stopped after ca. 300 documents, with the following message: `elasticsearch.helpers.BulkIndexError: 1 document(s) failed to index.` To me it seemed an overkill to let the whole process fail on this, so I wrapped the `streaming_bulk` operation in a try / except clause. Unfortunately, the error message is not more specific about which document failed and why. Curiously, locally the indexing operation finishes with 1960 documents having been indexed, i.e., all documents currently in this corpus.

### DIFF
--- a/backend/es/es_index.py
+++ b/backend/es/es_index.py
@@ -132,14 +132,17 @@ def populate(client: Elasticsearch, corpus: Corpus, start=None, end=None):
         settings.CORPUS_SERVER_NAMES.get(corpus_name, 'default')]
 
     # Do bulk operation
-    for success, info in es_helpers.streaming_bulk(
-        client,
-        actions,
-        chunk_size=corpus_server['chunk_size'],
-        max_chunk_bytes=corpus_server['max_chunk_bytes'],
-    ):
-        if not success:
-            logger.error(f"FAILED INDEX: {info}")
+    try:
+        for success, info in es_helpers.streaming_bulk(
+            client,
+            actions,
+            chunk_size=corpus_server["chunk_size"],
+            max_chunk_bytes=corpus_server["max_chunk_bytes"],
+        ):
+            if not success:
+                logger.error(f"FAILED INDEX: {info}")
+    except Exception as e:
+        logger.warning(e)
 
 
 def perform_indexing(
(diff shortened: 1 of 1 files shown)


lukavdplas:
> When indexing the JewishMigration corpus, the process stopped after ca. 300 documents, with the following message: elasticsearch.helpers.BulkIndexError: 1 document(s) failed to index. To me it seemed an overkill to let the whole process fail on this

It definitely sounds like overkill, but I think in a lot of situations, it's not inaccurate to say indexing is all-or-nothing. Some corpora just have to be complete in order to have research value. So in those cases, you don't want the indexing process to just keep on going after one document has already failed.

jgonggrijp:
@lukavdplas I agree that some corpora need to be complete. However, stopping the indexing as soon as one document fails will not help to eventually obtain a complete copy in the ES index.

BeritJanssen:
I agree. What's strange is that the document isn't missing at all once the indexing process is finished. I'll look into the `streaming_bulk` arguments to see, hoping we can get more information out of this function as well. I think `raise_on_error` should definitely be `True`, `raise_on_exception` may be `False`, as long as we have good output about the exception, allowing us to revisit and fix the issues causing the exception.

lukavdplas:
Hm, this may be a matter of personal preference.

However, something to consider: quite a few unit tests currently work on the assumption that if you complete indexing without raising an error, it means all documents were indexed. So right now, you can write a decent unit test by running an index job and assuming all documents are there if it completes. Also, many unit tests use test indices in some way, but the completeness of those indices is not asserted explicitly. 

So I think in a testing context, it would make sense to raise an error and quit if a document can't be indexed; i.e. add an option like `raise_on_exception` in the indexing function. If not, I would recommend going over existing unit tests for the `addcorpus`, `es`, and `corpora` modules, and make sure that they will actually break if the index is incomplete.

BeritJanssen:
I now experimented with `raise_on_exception=False` and `raise_on_error=False`, which seems to work nicely: the error is logged and gives some information of which document was problematic, and what the problem was. In this case, it was one field which did not have the correct format for the mapping. The document was still written to the index, without that field. I think, as long we have good output describing the problematic documents, this is a good solution.

PR #1660, opened 2024-09-20
fix: let es_index.create return versioned index name

### LINKED ISSUE #1656: Bug: cannot create new index / will write to wrong index version in production
Currently, it's not possible to index a new corpus in production as far as I can tell. This line seems to be the culprit:
https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/14c6817093a7d232c2c933a0dd9910fb22f44b4a/backend/es/es_index.py#L75

This used to be
```
corpus_definition.es_index = "{}-{}".format(
            corpus_definition.es_index, get_new_version_number(client, alias, corpus_definition.es_index))
```
which means the corpus_definition object was updated, so retrieving `corpus_definition.es_index` in the `populate` method would return the versioned index name.

I suspect that if we try to create a new version of an existing corpus with versioned index `myindex-1` and alias `myindex`, we can neatly create the `myindex-2` corpus, but write to `myindex-1` instead, as the `populate` method writes to the alias `myindex`, which still points to `myindex-1`.

Possible solutions:
- easiest: let `create` return the versioned index name, and pass this into to the `populate` method
- go back to `corpus_definition.es_index = versioned_index`, followed by `corpus_definition.save()`
- add a new field `es_index_version` to the CorpusConfiguration model, which is updated with the number returned by `get_new_version_number()` method upon creation

Disadvantage of the last two approaches: in case the index operation fails, the database will show a version number which does not correspond to the version the alias points to. This could be mitigated by decrementing the version number upon failure.

BeritJanssen:
For now, this PR should fix #1656 . Perhaps not the prettiest solution, but I wanted to avoid repeated requests to the Elasticsearch client just to reaffirm the current version number. Perhaps in the long run, some database bookkeeping of index versions would not be a bad idea, but let's discuss this at a later point.

### DIFF
--- a/backend/es/es_alias.py
+++ b/backend/es/es_alias.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import re
 
-from addcorpus.models import Corpus
+from addcorpus.models import Corpus, CorpusConfiguration
 from ianalyzer.elasticsearch import elasticsearch
 
 import logging
@@ -49,6 +49,13 @@ def alias(corpus: Corpus, clean=False):
     if len(actions) > 0:
         client.indices.update_aliases(actions=actions)
     logger.info('Done updating aliases')
+
+
+def get_current_index_name(corpus: CorpusConfiguration, client) -> str:
+    """get the name of the current corpus' associated index"""
+    alias = corpus.es_alias or corpus.es_index
+    indices = client.indices.get(index="{}".format(alias))
+    return max(sorted(indices.keys()))
 
 
 def get_new_version_number(client, alias, current_index=None):
--- a/backend/es/es_index.py
+++ b/backend/es/es_index.py
@@ -18,7 +18,7 @@ from addcorpus.models import Corpus, CorpusConfiguration
 from addcorpus.python_corpora.load_corpus import load_corpus_definition
 from addcorpus.reader import make_reader
 from ianalyzer.elasticsearch import elasticsearch
-from .es_alias import alias, get_new_version_number
+from .es_alias import alias, get_current_index_name, get_new_version_number
 import datetime
 
 import logging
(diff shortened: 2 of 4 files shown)


JeltevanBoheemen (approved):
I think this is a smart little solution to the problem.
Pending the index management functionality this has my approval.

For safety, I added a review reqest for @lukavdplas. I'm not that familiar with the indexing procedure. So any possible interactions this may cause come to mind?

BeritJanssen:
Ah, I'd asked you because you did the most recent changes in this file. Thanks for the approval, but let's see if @lukavdplas still spots anything we both may have missed indeed.

backend/es/es_index.py
```diff
@@ -93,20 +99,26 @@ def create(client: Elasticsearch, corpus: Corpus, add: bool = False, clear: bool
             settings=settings,
             mappings=es_mapping,
         )
+        return index_name
```
lukavdplas, line 102: If you're using the `--add` flag (to add documents to an existing index), this function won't reach this line, but `return None` on line 69. I expect that will cause an issue when `populate()` is called with `versioned_index_name = None`.

This should be straightforward to fix: just return the name of the existing index in that case.

lukavdplas (changes requested):
Nice, I think this solution makes a lot of sense.

I added one comment about using `--add`, which *I think* will not work in this version. Feel free to merge if I'm wrong about that. Otherwise, bonus points if you also add a unit test for that option :upside_down_face:

backend/es/es_index.py
BeritJanssen: Yes, and luckily, this is already caught by a unit test. :-)

backend/es/es_alias.py
```diff
@@ -51,6 +51,13 @@ def alias(corpus: Corpus, clean=False):
     logger.info('Done updating aliases')
 
 
+def get_current_index_name(corpus: CorpusConfiguration, client) -> str:
+    """get the name of the current corpus' associated index"""
+    alias = corpus.es_alias or corpus.es_index
+    indices = client.indices.get(index="{}".format(alias))
+    return max(sorted(indices.keys()))
```
lukavdplas, line 58: Is this sorting alphabetically? That would mean that if `my_corpus_9` and `my_corpus_10` both exist, you're use `my_corpus_9`, no?

PR #1667, opened 2024-09-27
Fix NER check

BeritJanssen:
Before, the existence of named entities was retrieved from the mappings. I decided to replace this by a search with a query to check if any fields starting with "ner:*" are actually in the documents. Less wordy this way, and with `size=0`, this request takes only 2-7 milliseconds to perform on the largest corpus we have at the moment (dutchnewspapers-all).

### DIFF
--- a/backend/addcorpus/models.py
+++ b/backend/addcorpus/models.py
@@ -1,5 +1,12 @@
 import warnings
 
+from django.contrib import admin
+from django.contrib.auth.models import Group
+from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.db.models.constraints import UniqueConstraint
+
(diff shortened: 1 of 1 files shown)


backend/addcorpus/models.py
```diff
@@ -264,14 +266,15 @@ def clean(self):
 
     @property
     def has_named_entities(self):
-        client = elasticsearch(self.es_index)
+        from es.search import total_hits
```
lukavdplas, line 269: Is there a reason this line isn't at the top of the file?

lukavdplas (approved):

backend/addcorpus/models.py
BeritJanssen: Yes, otherwise it's a circular import, as the `es.search` module imports `addcorpus.models.CorpusConfiguration`.

PR #1668, opened 2024-09-27
Format and transition numbers

oktaal:
Formats the hit count. Adds some transition effects. [Borrowed from MiMA](https://github.com/CentreForDigitalHumanities/mima/blob/78df25381d5ff892f905d02790ed4c50995cb2a2/frontend/src/app/transition-numbers.pipe.ts) 

showing the numbers

### DIFF
--- a/frontend/src/app/search/search-results.component.html
+++ b/frontend/src/app/search/search-results.component.html
@@ -1,13 +1,12 @@
 <ng-container *ngIf="pageResults?.result$ | async as page; else noResult">
     <div class="level">
-        <h2 class="subtitle" *ngIf="page.total <= 5">
-            {page.total, plural, =1 {1 hit} other {{{page.total}} hits}}
-        </h2>
-        <h2 class="subtitle" *ngIf="page.total > 5">
-            {{page.total}} results.
-            Showing results {{pageResults.from$ | async}}
+        <h2 class="subtitle">
+            {page.total, plural, =1 {1 result} other {{{page.total | transitionNumbers}} results}}
+            @if (page.total > 5) {
+            Showing results {{pageResults.from$ | async | transitionNumbers}}
             -
-            {{(pageResults.to$ | async)}}
+            {{(pageResults.to$ | async | transitionNumbers)}}
+            }
         </h2>
         <ia-highlight-selector *ngIf="!pageResults.highlightDisabled" [pageResults]="pageResults"></ia-highlight-selector>
     </div>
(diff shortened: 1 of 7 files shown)


frontend/src/app/search/search-results.component.html
```diff
@@ -1,13 +1,12 @@
...
-            Showing results {{pageResults.from$ | async}}
+        <h2 class="subtitle">
+            {page.total, plural, =1 {1 result} other {{{page.total | transitionNumbers}} results}}
+            @if (page.total > 5) {
+            Showing results {{pageResults.from$ | async | transitionNumbers}}
             -
-            {{(pageResults.to$ | async)}}
+            {{(pageResults.to$ | async | transitionNumbers)}}
+            }
         </h2>
```
lukavdplas, line 10: I've been meaning to make the hit count a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) - similar to how it's used on the [download tab](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/688b5f2eae122bf973aae45f8f03861ab19c208b/frontend/src/app/download/download.component.html#L2). I'm concerned that updating the numbers like this would spawn a barrage of updates in that case.

oktaal: Maybe the `aria-live="polite"` should be enough?

> "aria-live="polite" indicates that the screen reader should wait until the user is idle before presenting updates to the user. This is the most commonly used value, as interrupting the user with "assertive" might interrupt their flow." 

And there is also this interesting attribute [`aria-busy`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy)

> Used in [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), the global aria-busy state indicates an element is being modified and that assistive technologies may want to wait until the changes are complete before informing the user about the update.

lukavdplas: `aria-live=polite` basically means the message is queued and will be read the next time the screen reader is idle. But that's why I was worried this might queue a bunch of updates that are all read after one another.

But it looks like `aria-busy` would be the right solution here!

lukavdplas:
Btw, another question: could you use a media query to disable the animation with [prefers-reduced motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)?

This isn't strictly necessary, since this animation doesn't count as motion. However, some people (including myself) use the setting to just reduce the amount of animation on webpages.

PR #1669, opened 2024-10-02
Fix Peace Portal visualizations & add visualization for Jewishmigration

BeritJanssen:
I wanted to add visualizations for Jewishmigration and noticed in the process that the Peace Portal `field_defaults` used the old format for specifying visualizations. This branch should fix that.


lukavdplas (approved):
Changes look fine!

Please get your autoformatter under control, by the way.

PR #1673, opened 2024-10-03
Add basic registration ratelimit

ar-jan:
@tymees does this throttle implementation look reasonable?

Team, do we want `DEFAULT_THROTTLE_RATES` in common or local settings? And what's a rate that's high enough for potential peak registration periods (like a course/lecture)?


tymees:
The setup looks reasonable to me! 

The only thing I'd have to add is that the original CERT call also mentions a lack of throttling on the PW reset view (it's not very clearly written, but it is in there...), so you might want to apply something similar to that?

ar-jan:
Thanks, I missed that. Added.

BeritJanssen (approved):
Apart from the current message in the reset password template, looks good to me!

PR #1674, opened 2024-10-03
add ngram field to all corpora with date field and main_content_mapping

BeritJanssen:
All indices have been prepared for offering the "neighbouring words" visualization, but the corpus definitions themselves hadn't been updated to actually add those visualizations. This PR fixes that.


BeritJanssen:
Urgs. My apologies, my code formatter was still set to format on save. I adjusted that now, but these files are still heavily formatted. The changes basically come down to adding "ngram" in the following fields:
- Dutchnewspapers.content
- GuardianObserver.content
- Times.content
- Periodicals.content

In principle, we can probably also add this feature for corpora which have a `year` field, instead of a `date` field, but I left those alone for now.

JeltevanBoheemen:
The formatting makes it hard to see what's going on, but just looking at the changes you pointed out: looks good!

JeltevanBoheemen (approved):

PR #1677, opened 2024-10-10
Always add term vector to main content field and fix mappings in some parliamentary corpora

BeritJanssen:
This PR removes the argument `updated_highlighting` from the `main_content_mapping`, as we have updated all indices now with the term vector, and this should be standard behaviour.

The branch also fixes some issues with parliamentary corpora: `finland` didn't yet set a language, and `ireland` overrode the es_mappings in a way that's not necessary anymore when not passing a `language` argument to `speech()`.


Meesch (approved):
Looks good!

PR #1678, opened 2024-10-10
Overhaul GitHub actions once more

BeritJanssen:
This branch reworks the current strategy of GitHub actions, which often fail due to failed pushes to the registry. It tries to limit the pushes to the registry with the following strategy:
- every push to `feature/**`, `bugfix/**`, `hotfix/**`, `release/**`, `develop` and `master` will trigger a test action, using the `:latest` image of frontend / backend / elasticsearch, respectively.
- in dependabot/feature/bugfix/hotfix branches: when `requirements.txt` or `yarn.lock` have been touched, another action will be triggered: we will run tests *without* using the image, this to avoid failing tests due to missing / outdated libraries. Unfortunately I cannot see how I would then avoid still triggering the above action, as it does not seem we can tell GHA to run actions "if this path is changed, but absolutely not if this other path is changed".
- when merging PR's with changes to requirements or yarn.lock to develop, a new image with :latest tag will be built and pushed
- every month, a new set of images for frontend, backend and elasticsearch will be generated and pushed to the `:latest` tags

Does this make sense, @tymees ?

I'm also wondering whether to use build stages after all in the Dockerfiles, then the "scheduled" workflow could more specifically target only the part of the images prior to installing with `pip` or `yarn`, but then I'm not sure how I could get the other actions to build on top of that image.

Another open question: does the `cache-to` and `cache-from` in the actions actually still make sense? I *do* sometimes see `CACHED` in the action reports, but I'm not sure when this would actually happen, as it seems every time an action builds, it will actually push to a new hashed version of the image.

### DIFF
--- a/.github/workflows/frontend-build-and-push.yml
+++ b/.github/workflows/frontend-build-and-push.yml
@@ -8,6 +8,7 @@ on:
         - closed
       paths:
         - frontend/yarn.lock
+        - 'docker-compose.yaml'
 
 jobs:
   if_merged:
--- a/.github/workflows/frontend-build-and-push.yml
+++ b/.github/workflows/frontend-build-and-push.yml
@@ -1,6 +1,7 @@
 name: Frontend build and push after merge of yarn.lock
 
 on:
+    workflow_dispatch:
     pull_request:
       branches:
         - develop
--- a/.github/workflows/frontend-build-and-push.yml
+++ b/.github/workflows/frontend-build-and-push.yml
@@ -1,0 +1,34 @@
+name: Frontend build and push after merge of yarn.lock
+
+on:
+    pull_request:
+      branches:
+        - develop
+      types:
+        - closed
+      paths:
+        - frontend/yarn.lock
+
+jobs:
+  if_merged:
+    name: Build and push frontend image
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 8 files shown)


.github/workflows/frontend-build-and-push.yml
```diff
@@ -0,0 +1,35 @@
...
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build frontend image, using cache from Github registry
+      uses: docker/build-push-action@v6
+      with:
+        context: frontend/.
+        push: true
+        tags: ghcr.io/uudigitalhumanitieslab/ianalyzer-frontend:latest
```
tymees, line 33: Seeing as this repo is now in the CDH org, does it make sense to keep pushing to the old organisation's docker registry?

tymees (commented):
I think this makes sense yes;

As a suggestion, you might want to move the scheduled rebuild of frontend/backend into the `{x}-build-and-push` workflows (you can have both triggers in there at the same time). That way, you only have one workflow with the build-steps (reducing duplication). However, as the gains are minimal, I'd say it might not be worth it :)

> I'm also wondering whether to use build stages after all in the Dockerfiles, then the "scheduled" workflow could more specifically target only the part of the images prior to installing with pip or yarn, but then I'm not sure how I could get the other actions to build on top of that image.

I'd say that that's a bit over-optimizing, a few minutes of extra work once a month really isn't going to matter much. 
In addition, you might actually want to test if pip/yarn can still run properly with the changed system-packages?

> Another open question: does the cache-to and cache-from in the actions actually still make sense? I do sometimes see CACHED in the action reports, but I'm not sure when this would actually happen, as it seems every time an action builds, it will actually push to a new hashed version of the image.

I'm also not entirely sure; I'd say they are less likely to matter if you don't build as often as before, so in that sense they might be redundant. I'd have to look deeper into docker-caching to be sure tho.

.github/workflows/frontend-build-and-push.yml
BeritJanssen: Good catch! Will investigate.

PR #1679, opened 2024-10-11
Test throttle rates

ar-jan:
WiP, some things to resolve here still:

The throttle rate from the fixture is not actually applied, although its value is read correctly in the test function (same for `@override_settings`). The settings that are actually used are the defaults from `common_settings.py`.
I thought I might have to use the `rest_framework.test APIClient` for this, but it seems to behave the same whether I use the default `client` or `drf_client` I added here.

<details><summary>cache issue</summary>
In addition, this is strange: when I test manually with `'registration': '1/minute'`, the first registration succeeds and the second one gets a 429, as expected. But with the default rate of `'registration': '5/minute'` that is (apparently?) picked up by pytest, we get this:

> Request 1 status: 201
> Request 2 status: 201
> Request 3 status: 201
> Request 4 status: 201
> Request 5 status: 429
> Request 6 status: 429

I would've expected Request 5 to still succeed. Caching issue?
</details>


ar-jan:
The last point was indeed due to the cache from the first registration test.

backend/users/tests/test_registration.py
```diff
@@ -46,3 +52,44 @@ def test_register_verification(client, db, django_user_model):
     assert allauth_email.email == creds.get('email')
     assert allauth_email.verified is True
     assert allauth_email.primary is True
+
+
+def test_register_throttling(client, throttle_settings):
```
lukavdplas, line 57: > The throttle rate from the fixture is not actually applied, although its value is read correctly in the test function

Perhaps an obvious suggestion, but did you try this?

```suggestion
def test_register_throttling(throttle_settings, client):
```

ar-jan: I've now tried this, but the result is the same.

JeltevanBoheemen (approved):
This works fine now!

PR #1692, opened 2024-11-06
Add Le Figaro corpus definition

BeritJanssen:
Related to https://github.com/CentreForDigitalHumanities/I-analyzer/issues/1089, this branch adds a general `gallica` corpus definition, as well as a subcorpus, Le Figaro.

Right now, still has custom requirements, will be adjusted once [the corresponding branch in ianalyzer-readers](https://github.com/UUDigitalHumanitieslab/ianalyzer-readers/pull/26) is merged and released.

NB: failing test is using the existing Docker image, which doesn't have the ianalyzer-readers update. So that's as expected.


backend/corpora/gallica/images/figaro.png
lukavdplas, line 1: Won't this crop to "ga" in the interface?

lukavdplas: By the way, this image might be a good alternative? https://commons.wikimedia.org/wiki/File:Mary_Cassatt_Reading_Le_Figaro.jpg

backend/corpora/gallica/gallica.py
```diff
@@ -0,0 +1,124 @@
...
+            csv_core=True,
+        )
+
+    def identifier(self):
+        return FieldDefinition(
+            name="id",
+            display_name="Publication ID",
+            description="Identifier of the publication on Gallica",
+            es_mapping=keyword_mapping(),
+            extractor=XML(Tag("dc:identifier"), transform=lambda x: x.split("/")[-1]),
```
Meesch, line 102: I think this might result in an IndexError if it cannot split the contents of the `dc:identifier` element.

Meesch (approved):
Two minor things in the comments, looks good! I am excited to see this implementation of a generalized corpus definition for a set of corpora instead of a folder with utils.

backend/corpora/gallica/gallica.py
```diff
@@ -0,0 +1,124 @@
...
+            f"{self.data_url}/services/Issues?ark=ark:/12148/{self.corpus_ark}/date"
+        )
+        year_soup = BeautifulSoup(response.content, "xml")
+        years = [
+            year.string
+            for year in year_soup.find_all("year")
+            if int(year.string) >= start.year and int(year.string) <= end.year
+        ]
+        for year in years:
+            response = requests.get(
```
JeltevanBoheemen, line 44: This is half question half comment:

What happens in case of server being unresponsive, internet connection failing for a split second, stuff like that?
In the ideal case harvesting and indexing would be split operations so one failing does not lead to starting over completely. Unsure ho wwe approached this in previous API-exposed corpora, hence the question.

JeltevanBoheemen: Discussed this a bit with Luka, and this is how we usually approach online corpora. So ignore the comment!

backend/addcorpus/validation/indexing.py
```diff
@@ -87,12 +87,19 @@ def validate_has_data_directory(corpus):
...
-            'Missing data directory'
-        )
+    if not config.data_directory and not config.data_url:
+        raise CorpusNotIndexableError('Missing data directory or url')
 
-    if not os.path.isdir(config.data_directory):
-        raise CorpusNotIndexableError(
-            'Configured data directory does not exist.'
-        )
+    if corpus.data_dircetory and not os.path.isdir(config.data_directory):
```
lukavdplas, line 93: @BeritJanssen note the typo here (hence why the backend tests are failing)

lukavdplas:
@BeritJanssen I've reverted the merge commit because it's causing errors that are interfering with other development. Can you fix them and then re-open the PR?

These errors are caused by issues in `backend/addcorpus/validation/indexing.py`. See the [test output](https://github.com/CentreForDigitalHumanities/I-analyzer/actions/runs/11839057596/job/32989612698).

The test output reports a typo, but the same tests will also fail if you correct it. The lines that follow it seem to expect that the `corpus` argument is a `CorpusDefinition` (i.e. the custom Python class defining the corpus), rather than a `Corpus` (i.e. the database object). 

However, these lines are run after verifying that the corpus does _not_ have a `CorpusDefinition` class, which also suggests a deeper issue with what the code was meant to do. (This makes the fix less straightforward; hence why I reverted the merge instead of pushing a quick fix.)

The reason why this validation is skipped for corpora with a Python definition, is that it functions as a pre-check for reading source files. But since Python corpora implement a custom script for that, it's hard to set universal expectations.

In my view, the validation logic that is added here describes a decent default, but will cause issues if it has no options for customisation. It includes some assumptions that may not be universal, like `data_url` using HTTP protocol.

If we do want to enforce these conditions, they should at least be documented clearly. The [current documentation](https://github.com/CentreForDigitalHumanities/I-analyzer/blob/develop/documentation/Writing-a-corpus-definition-in-Python.md) describes how to implement `sources()` and `data_directory` but does not suggest the kind of restrictions that are implemented here.

PR #1693, opened 2024-11-06
Add search filter and visualization for euparl speaker_country

BeritJanssen:
Per request by the researcher team, this branch adds a search filter and visualizations for the `speaker_country` field.

### DIFF
--- a/backend/corpora/parliament/euparl.py
+++ b/backend/corpora/parliament/euparl.py
@@ -11,6 +11,7 @@ from ianalyzer_readers.extract import Backup, Combined, Metadata, RDF
 
 from addcorpus.es_mappings import keyword_mapping
 from addcorpus.python_corpora.corpus import FieldDefinition, RDFCorpusDefinition
+from addcorpus.python_corpora.filters import MultipleChoiceFilter
 from corpora.parliament.parliament import Parliament
 import corpora.parliament.utils.field_defaults as field_defaults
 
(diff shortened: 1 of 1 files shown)


backend/corpora/parliament/euparl.py
```diff
@@ -193,15 +194,18 @@ def parse_graph_from_filename(self, filename: str) -> Graph:
     )
 
     speaker_country = FieldDefinition(
-        name='speaker_country',
-        display_name='Represented country',
-        description='The EU country the speaker represents',
+        name="speaker_country",
```
JeltevanBoheemen, line 197: Probably autoformatter deciding to go with double quotes?

JeltevanBoheemen (approved):

backend/corpora/parliament/euparl.py
BeritJanssen: Yes, indeed.

lukavdplas: At present, `'` is more commonly used in the I-analyzer backend, so if your reason for using an autoformatter is consistency, it probably makes more sense to configure the formatter with single quotes.

BeritJanssen: Yes, problematic is that for another project I'm using double quotes now. :-P

BeritJanssen: The auto-formatter is "Black", and overall I'm quite happy with how it works. I will check if I can configure it otherwise.

PR #1694, opened 2024-11-07
Add Estimated Centuries filter in Jewish Migration

BeritJanssen:
This branch makes sure that the "Estimated Centuries" field can be filtered on, useful to compare the map over time.

Screenshot 2024-11-07 at 16 19 25

Note: for this to work properly, will need to reindex the JewishMigration corpus once more, to convert from integer to keyword field.


tijmenbaarda (approved):
Looks great! I assume that no changes are needed in the Django jewishmigration application?

BeritJanssen:
@tijmenbaarda : no changes on the Django jewismigration app needed indeed.

Issue #1696, opened 2024-11-08
IndexJob model


lukavdplas:
This is part of #1007 - building a system to handle indexing without a command line.

Beside representing the actual index (#1695), it will be useful to keep track of index-related actions. We can create an `IndexJob` model for this, which represents any kind of "job" related to indices. Functions:

- Keep a log of indexing tasks. Useful when you have multiple developers and multiple projects on the same server - which we do.
- Allow planning, viewing, and starting jobs through the admin site.
- Retain data so index jobs can be queued, report on their progress, etc.

The actual contents of the job can be broken up into related `IndexTask` objects, e.g. `CreateIndexTask`, `PopulateIndexTask`, `AddAliasTask`, etc. To run a job, you run each of the tasks related to it.

Goals for this issue:

- Create `IndexJob` models (and related task models).
- When indexing from the command line, save an `IndexJob` to log the action.
- Show the IndexJob as read-only in the admin site.

PR #1698, opened 2024-11-08
Feature/index overview

### LINKED ISSUE #1695: Represent elasticsearch index in postgres database
This is part of #1007 - building a system to handle indexing without a command line.

Ultimately, the issue to be solved is that we don't have a good way to track what indices exist and what's happening with them.

Proposed feature: Add database models to represent indices. The models just needs to be a pointer to a name on a cluster that other models can refer to.  Additional metadata (e.g. mappings) can be fetched from Elasticsearch.

It will be useful to add an "index discovery" command where I-analyzer will fetch the names of all indices it has access to.

However, it's also useful to refer to indices that aren't currently available. For example, you don't want the server to erase all references to an index whenever it can't connect to the Elasticsearch cluster. Also, creating a reference for an index that doesn't exist yet can be a useful way to plan index-related tasks.

lukavdplas:
Adds tables for Elasticsearch clusters and indices to the database. Close #1695

The models are not *used* anywhere yet, but will be useful in #1696. For now, the immediate gain is that admins can see an overview of available indices in the admin site. This isn't anything you can't see in Kibana, but it's neat.

screenshot of i-analyzer admin site, showing metadata about an index

Keeping the models in sync is all optional at this point, but can be done as follows:

- There is a django-admin command, `update_index_metadata`, which will update servers from the Django settings, and indices for each server. I've added this to the `yarn start-back` command.
- The admin site includes an action to update index metadata per server, or update the availability status of a single index.

### Changes for deployment

It's not strictly necessary to change anything, but I recommend adding the `update_index_metadata` command to the startup routine in the deployment script.


backend/es/admin.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def has_change_permission(self, request, obj=None):
+        # disable editing
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        # only inactive servers can be deleted
+        if obj and obj.active:
+            return False
+        return super().has_delete_permission(request, obj)
```
JeltevanBoheemen, line 21: Good thinking!

backend/es/migrations/0002_alter_index_available.py
JeltevanBoheemen, line 1: Prefer wrapping these two migrations into one before merging

backend/es/models.py
```diff
@@ -0,0 +1,169 @@
...
+        Try to connect to the Elasticsearch server.
+
+        Returns a boolean, indicating if the connection succeeds.
+        '''
+        client = self.client()
+        if not client:
+            return False
+        try:
+            client.info()
+        except:
```
JeltevanBoheemen, line 66: May want to raise, or at least log, something here.

JeltevanBoheemen (approved):
Very thorough and clear implementation. I'm fucking impressed. Only slight worry is the new commands failing silently. Anything to do with inter-server connections has a real possibility to fail, that we may want to account for.

PR #1701, opened 2024-11-14
Change indexing logger to propagate: False

ar-jan:
When indexing a corpus, logging messages are duplicated in the console. The root logger logs to console, so the indexing logger should be `"propagate": False`.


lukavdplas (approved):
Ah, I was wondering why the logs were duplicated. Thanks for figuring that out! :+1:

PR #1703, opened 2024-11-15
resolve warnings

lukavdplas:
This resolves a lot of the backend warnings that were raised during testing or runtime.

Most of these changes are very tame. The only point of concern is the library change for manipulating PDFs, as the affected modules have insufficient test coverage. The change in backend requirements is also causing the failing CI check, but from what I understand, the build does actually succeed and is just reported in a confusing way.

Changes:

- Migrated from `PyPDF2` to `pypdf` library (PyPDF2 is deprecated)
- Updated some fields in mock corpora to use recommended configuration
- Updated depecrated parameters to some `Elasticsearch` method calls
- Removed obsolete `USE_L10N = True` setting from Django project settings.
- Use timezone-aware input when filtering a user's recent queries
- Use raw strings for some regular expressions
- Renamed a function that started with `test_` but was not a unit test.
- During test setup, ignore warnings about test corpora not having an "ID" field.
- During test setup, ignore warnings about mapping types in the Times corpus. Those warnings are still raised in the tests in `corpora/times`, but are not relevant elsewhere.
- Added test specifications for `source_archive` field in `parliament-netherlands` corpus
- In the ECCO and Delpher corpora, removed the `searchable=True` setting from some fields. (Because these fields lack the appropriate mapping, you can't search them successfully at the moment.)


BeritJanssen (approved):
Excellent!

PR #1708, opened 2024-11-18
Hotfix/field values

### LINKED ISSUE #1705: Numeric/boolean field values not displayed
### What went wrong?

In the document view, fields with integer, float, or boolean values show blank values, when the data does have a value. I think this does not *always* happen but I'm not sure what triggers it. Maybe if the source data provided a string instead of an integer/float?

The values are not actually blank in the underlying data, so they should be displayed.

### Screenshot

screenshot of document in the DBNL corpus in I-analyzer

Note the blank values for "content available" and "chapter index". But we know that all documents have a value for this field:

screenshot of fields information in I-analyzer, displaying information about the "content available" field. This stats that 100% of documents in the corpus have a value for the field.

- [ ] a local server

### Version

5.14.0

### Steps to reproduce

View a document with numeric/boolean fields, e.g. https://ianalyzer.hum.uu.nl/document/dbnl/_tij008186901_01_0055

You will also encounter the same issue in document previews in the search results. This is caused a different underlying function which runs into the same error (the highlight pipe rather than the paragraph pipe, I think), where it attempts to use a string method on a non-string. But I don't think we have any corpora online which show boolean/numeric values in the preview. So steps to reproduce would be:

- Open a local copy for a corpus with numeric/boolean fields, e.g. DBNL
- Adjust the field configuration so those fields are included in document previews
- Open the search page for the corpus.

### LINKED ISSUE #1706: Document preview shows ellipses when content is not truncated
### What went wrong?

Document previews will show ellipses (`...`) for fields, even when the content is actually displayed in full.

If the content is not truncated, the preview should not add ellipses.

### Screenshot

screenshot of document in I-analyzer troonredes corpus

- [ ] a local server

### Version

5.14.0

### Steps to reproduce

Go to https://ianalyzer.hum.uu.nl/search/troonredes

lukavdplas:
Fixes #1705, #1706, #1709

This also creates two separate components for displaying fields, instead of writing that logic in the document-view and document-preview components, because I was getting lost in the templates of those components.

I've been meaning to clean up these components a bit and straighten out some minor inconsistencies. For now, this at least fixes the missing values.

### DIFF
--- a/frontend/src/app/document/content-field-preview/content-field-preview.component.html
+++ b/frontend/src/app/document/content-field-preview/content-field-preview.component.html
@@ -1,0 +1,17 @@
+<ng-container *ngIf="field && document">
+    <ng-container *ngIf="document.highlight && document.highlight[field.name];
+        else unhighlightedRow">
+        <ng-container *ngFor="let highlight of document.highlight[field.name];
+            last as isLast">
+            <div ngPreserveWhitespaces style="word-break:break-word"
+                [innerHtml]="highlight">
+            </div>
+            <hr *ngIf="!isLast">
+        </ng-container>
+    </ng-container>
+    <ng-template #unhighlightedRow>
+        <div style="word-break:break-word"
+            [innerHtml]="document.fieldValue(field) | snippet">
+        </div>
+    </ng-template>
+</ng-container>
--- a/frontend/src/app/document/metadata-field/metadata-field.component.html
+++ b/frontend/src/app/document/metadata-field/metadata-field.component.html
@@ -1,0 +1,11 @@
+<ng-container *ngIf="field">
+    <div *ngIf="!isUrlField(field) && !isGeoPointField(field)"
+        [innerHtml]="field | elasticsearchHighlight:document | paragraph">
+    </div>
+    <ng-container *ngIf="isGeoPointField(field)">
+        {{field | geoData:document}}
+    </ng-container>
+    <ng-container *ngIf="isUrlField(field)">
+        <a href={{document.fieldValue(field)}} target="_blank">{{document.fieldValue(field)}}</a>
+    </ng-container>
+</ng-container>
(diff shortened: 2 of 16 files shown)


frontend/src/app/document/content-field-preview/content-field-preview.component.html
```diff
@@ -0,0 +1,17 @@
+<ng-container *ngIf="field && document">
+    <ng-container *ngIf="document.highlight && document.highlight[field.name];
+        else unhighlightedRow">
+        <ng-container *ngFor="let highlight of document.highlight[field.name];
+            last as isLast">
```
JeltevanBoheemen, line 5: Lots of `ngIf` and `ngFor` here. I wonder if an elvis operator could cut some lines?

JeltevanBoheemen: E.g. something like `*ngFor=let highlight of document?.highlight[field?.name]`
Probably wishful thinking :)

frontend/src/app/document/content-field-preview/content-field-preview.component.html
```diff
@@ -0,0 +1,17 @@
...
+            <div ngPreserveWhitespaces style="word-break:break-word"
```
JeltevanBoheemen, line 6: Minor `DRY` for the `style`. Could make it a class.

frontend/src/app/document/metadata-field/metadata-field.component.html
```diff
@@ -0,0 +1,16 @@
+<ng-container *ngIf="field" [ngSwitch]="field.displayType">
+    <div *ngSwitchCase="'text'"
+        [innerHtml]="field | elasticsearchHighlight:document | paragraph">
```
JeltevanBoheemen, line 3: This line is unclear to me. What is displayed as `innerHtml` in which cases?

JeltevanBoheemen (approved):
Nice. Good idea to declutter `DocumentView`. Some questions/comments.

lukavdplas:
I agree with your comments here; this is the kind of stuff I'm talking about when I say I still want to clean up these components some more. I think I'll make an update for that soon; I'll try to address your suggestions there as well.

PR #1711, opened 2024-11-20
image licensing

lukavdplas:
This aims to address a few copyright-related issues:

- I-analyzer reuses several images with a CC BY or CC BY-SA licence without providing attribution, so our reuse is not covered by the licence.
- Sharing this entire repository under an MIT licence also violates those licences.
- Most images have no clear attribution or source, so even if they are in the public domain, this is a pain in the ass to verify.

Solution:

- Add a `*.license` file next to each image in the repository and mention this in the readme.
- Mention attribution and licence in the corpus documentation, so it's also visible for users.

I've added licence info for all corpora where I could find the source of the image. Here is an overview of ones I could not trace:

- [ ] **Dutch Annual Reports:** can't find a source for this - it is (or in some cases, was) available on some stock photo platforms, but I could not find any page that claimed to be the original author.
- [ ] **ECCO:** This image is often used around the ECCO corpus, for example in [this description from TCP](https://textcreationpartnership.org/tcp-texts/ecco-tcp-eighteenth-century-collections-online/). I assume that it's cropped and edited from an illustration that is in the public domain, and it's likely that no one would claim copyright over the editing. But I can't find an actual source for it.
- [ ] **DIOPTRA-L:** Clearly designed for this project, but no licence or credit is included. I expect that we have specific permission to use it without attribution but there is no general licence.
- [ ] **Jewish Inscriptions / Jewish Migration:** Reverse image search turns up nothing. May be the same situation as DIOPTRA-L, though as with the ECCO corpus, this may be considered public domain.
- [ ] **People & Parliament: France:** Can't find a source for this. Reverse image search is unproductive because there are many similar pictures of the same room.
- [x] **People & Parliament: Ireland:** This image was [originally intended as temporary](https://github.com/CentreForDigitalHumanities/I-analyzer/pull/1025) due to copyright. Replaced the image with one that has a CC licence.
- [ ] **Troonredes:** Can't find the source for this one. It's probably less effort to just find a suitable replacement.
- [ ] **U-Blad:** No source provided. This image is UU-specific, so again, I expect we have permission but there is no licence. Note that it is a scan of copyrighted works, too.


BeritJanssen:
With ECCO, the image was supplied with the source data, as far as I remember. As for Dutch Annual Reports - I must've picked that from an image search for images in the open domain, but if that's not something we can reproduce, probably easier to go with a new image. Perhaps of the Amsterdam Zuidas, so it's more recognizably Dutch? DIOPTRA-L: I suspect that this image was produced by Alex back then, I'll check with Haidee if she knows. The image from JewishMigration / Inscriptions is from before my time. Again, I don't think anyone would be outraged if we picked a different (comparable) image here. For France, I think we can just go for any other image of the same room, too.

jgonggrijp:
I didn't mean to suggest that you should include the copyright and license information only in the metadata and nowhere else. You should definitely mention the fact that some images have separate licenses as well elsewhere, for example in the README (and in the corpus documentation as you suggested). This will alert people to whom it matters to look for those metadata. The metadata are just an alternative for the `.license` files.

jgonggrijp:
More files more clutter. It's just a suggestion though, you can ignore it if you don't like it.

BeritJanssen:
Dutch Annual Reports: https://commons.wikimedia.org/wiki/File:RotterdamMaasNederland.jpg
ECCO: https://commons.wikimedia.org/wiki/File:English_humorists_of_the_eighteenth_century_-_Sir_Richard_Steele,_Joseph_Addison,_Laurence_Sterne,_Oliver_Goldsmith_(1906)_(14802513503).jpg
French Parliament: https://commons.wikimedia.org/wiki/File:Hemicycle_Senat_France.jpg

Would this be suitable replacements?

lukavdplas:
Great, thanks for looking into this! :)

PR #1717, opened 2024-12-02
add creation date to corpora

### LINKED ISSUE #1097: Add creation dates to corpora
Save the date on which a corpus was first added to the database.

This isn't immediately useful since I imagine we will just let all existing corpora default to 'today'.

Future uses:
- When researchers can add their own corpora, this can help us get an overview
- We can use it to highlight new corpora in the corpus overview ✨

lukavdplas:
close #1097

Corpora are now sorted most to least recent on the home page. However, the creation date for an existing corpora will just be the date on which the migration is run, so this will only affect new corpora.


backend/addcorpus/models.py
```diff
@@ -57,6 +57,10 @@ class Corpus(models.Model):
         help_text='whether the configuration of this corpus is determined by a Python '
             'module (some features are only available for Python-based corpora)',
     )
+    date_created = models.DateField(
+        auto_now_add=True,
+        help_text='date on which the corpus was added to the database',
+    )
```
BeritJanssen, line 64: You could also do:
```suggestion
class Meta:
        ordering = ["date_created"]
```
instead of explicitly calling `order_by` in the `views.py`. Or do you want a different ordering in different situations?

BeritJanssen, line 64: I see that I cannot actually make a suggestion on unchanged code, but below on the `class Meta:` definition, you could add:
```python
ordering = ["date_created"]
```
instead of explicitly calling `.order_by` in `views.py`. Or are you after being able to sort according to different criteria?

BeritJanssen (approved):

backend/addcorpus/models.py
lukavdplas: No, but I didn't think that ordering by date was relevant in other contexts.

PR #1722, opened 2024-12-04
Feature/uu course descriptions

lukavdplas:
Adds several corpora with UU course descriptions.

I've kept the old "humanities course descriptions" corpus, because that one uses a slightly different data format. That corpus is kept the same, but uses the same picture as the "updated" version.

Because these corpora are created for a separate server, there are a few minor changes to the frontend to configure the server in `environment.ts`:
- Added configuration for the brand (logo / title) shown in the navbar
- You can hide the corpus filters on the home page. This makes sense if, like here, the corpora on the server are very uniform so there is nothing to filter on.

## Server configuration changes

- If a server was using the `hum_course_descriptions` corpus, it must update the location of the source file. It's now located in the `uu_course_descriptions` module.
- Servers must update the `environment.ts` file to specify the `navbarBrand` setting.


frontend/src/app/menu/menu.component.ts
```diff
@@ -40,6 +41,14 @@ export class MenuComponent implements OnDestroy, OnInit {
         private route: ActivatedRoute
     ) {}
 
+    /**
+     * whether the brand should explicitly add "i-analyzer" to the app name; not
+     * necessary if the app name already includes "i-analyzer"
+     */
+    get showIAnalyzerInBrand() {
+        return ! this.appName.toLowerCase().includes('i-analyzer');
```
Meesch, line 49: I like the solution stylistically, but I don't know if this is the best way to implement naming conventions. I would either add some documentation explaining how the naming of apps works and how a developer should use those values, or just give two fields in the environment for Main app name and Secondary App name or something? Or is this meant to enforce that we have the I-analyzer name present at all times? :p

Meesch (changes requested):
Looks very nice, especially the landing page with all the faculties: great job! Got some small stuff and some comments, I tested only using the web app on the server.

frontend/src/app/menu/menu.component.ts
lukavdplas: Fair point, a primary / secondary name would make more sense!

backend/corpora/uu_course_descriptions/uu_course_descriptions.py
```diff
@@ -0,0 +1,337 @@
...
+            visualizations=['resultscount', 'termfrequency'],
+        ),
+        FieldDefinition(
+            name='department_description',
+            display_name='Department',
+            description='Coordinating department for the course',
+            extractor=CSV('OMSCHRIJVING'),
+            es_mapping=keyword_mapping(enable_full_text_search=True),
+            results_overview=True,
+            search_filter=MultipleChoiceFilter(option_count=100),
```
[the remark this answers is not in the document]
lukavdplas: It does? At least for me:

screenshot of "department" filter in I-analyzer

The CSV column name is misleading, though, maybe you missed it?

PR #1727, opened 2024-12-12
Clean up routing behaviour in ngram component

### LINKED ISSUE #1268: ngram request triggered three times
When trying to see the "neighbouring words" graph and opening the Network inspector, this shows the ngram request to be triggered three times. The call stack shows the following:
- first trigger `onChanges`. We listen to the corpus and queryModel input here. The initial state of the data does (but should not) trigger `loadGraph`.
- second trigger: `setStateFromParams`
- third trigger: `setStateFromParams`
One of the two triggers is likely caused by the updates from the `visualization.component` (see also #1374).

BeritJanssen:
Close #1268. As discussed IRL, I decided to focus on the problems involved in managing routing parameters connected to the ngram component, and this fixes the multiple fetching. My personal preference is to keep the concerns for parameter changes and requests separated, i.e., not to combine them both in *one* model. But extending `SongSync` and using its methods works like a charm! 🤩


frontend/src/app/models/field-filter.spec.ts
```diff
@@ -1,9 +1,5 @@
-import { convertToParamMap } from '@angular/router';
 import { mockFieldMultipleChoice, mockFieldDate, mockField } from '../../mock-data/corpus';
-import { EsDateFilter, EsTermsFilter } from './elasticsearch';
 import { BooleanFilter, DateFilter, DateFilterData, MultipleChoiceFilter } from './field-filter';
-import { of } from 'rxjs';
-import { distinct } from 'rxjs/operators';
 import { SimpleStore } from '../store/simple-store';
 import { Store } from '../store/types';
```
BeritJanssen, line 4: This change removes some unused imports, which I noticed when I was looking at other examples for testing routing behaviour.

frontend/src/app/models/visualization.spec.ts
BeritJanssen, line 1: The only test in this spec was the one related to the ngram related models and types.

frontend/src/app/visualization/ngram/ngram.component.spec.ts
```diff
@@ -38,32 +55,54 @@ describe('NgramComponent', () => {
...
+  it('should initialize ngramParameters with default values', () => {
+    expect(component.ngramParameters.state$.value).toEqual(defaultSettings);
+  });
+
+  it('should not abort tasks when `onParameterChange` is triggered during initialization', () => {
+    spyOn(component.stopPolling$, 'next');
+    component.onParameterChange('size', 2);
+    expect(component.stopPolling$.next).not.toHaveBeenCalled();
+  })
+
```
BeritJanssen, line 75: On initialization, all values in the template will be adjusted to the values reflected in the routing parameters, which will trigger the `onParameterChange` method, hence the extra `if` condition in there to see if the parameter *has* actually changed.

frontend/src/app/visualization/ngram/ngram.component.ts
```diff
@@ -39,6 +39,8 @@ export class NgramComponent extends ParamDirective implements OnChanges {
 
     @ViewChild('chart-container') chartContainer: ElementRef;
 
+    keysInStore = ['ngramSettings'];
+
```
lukavdplas, line 43: I think this variable is never used?

frontend/src/app/models/ngram.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        return {
+            size: 2,
+            positions: 'any',
+            freqCompensation: false,
+            analysis: 'none',
+            maxDocuments: 50,
+            numberOfNgrams: 10,
+            dateField: 'date'
```
lukavdplas, line 56: The way the date field is handled looks a bit shaky, since the corpus may not actually have a field with this name. I think this was already an issue, though. Have you checked that the visualisation still works with corpora like parliament-denmark-old?

lukavdplas (approved):
Looks good!

I wasn't sure about the way date fields are set to a default value here, but if that works, feel free to merge.

frontend/src/app/visualization/ngram/ngram.component.ts
BeritJanssen: Yes, by the `StoreSync` parent.

frontend/src/app/models/ngram.ts
BeritJanssen: Yes, good point. The problem is that when the `ngramParameters` model is initialized, the `corpus` input hasn't been set yet, so we cannot set the date field. Perhaps it would break functionality least if I excluded the date field from the route parameters for now. This is exactly why I was wondering how realistic it is to transfer data conversion logic to models, as we will also have to transfer knowledge about possible states in the component to the model.

BeritJanssen: Hmm, I tried and abandoned the idea to initialize `ngramParameters` in `ngOnChanges`. That seems to work in practice, but unit testing depending on change detection is a bit of a nuisance. I decided to remove the date field from the routing parameters for now: the number of corpora in which different date fields are available is so limited that I think it's not too important to include the date field in the route at this moment.

lukavdplas: Besides the issues in unit testing, it makes sense to me to create the parameters model after you have received all relevant input.

I've been looking into providing the `Corpus` through a resolver instead of input, which may be more appropriate for something like this. That would have the side effect that you could already access the corpus earlier in the component lifecycle.

PR #1729, opened 2024-12-19
Add documentation about fuzzy word search

### LINKED ISSUE #1712: Documentation: add more information on Elasticsearch query syntax to the manual
The option to search for words in a context of *n* words with an Elasticsearch query (see https://github.com/CentreForDigitalHumanities/I-analyzer/issues/1648) is not documented yet, but should be, even if we may offer an advanced search UI at some point to facilitate those queries.

BeritJanssen:
close #1712 .

### DIFF
--- a/frontend/src/assets/manual/en-GB/query.md
+++ b/frontend/src/assets/manual/en-GB/query.md
@@ -27,6 +27,18 @@ By default the search will combine all terms using `OR`. This means that when yo
 ### Be Careful with Spaces
 Adding or removing a space can change the results of your query. For example search for `+- term` is different than searching for `+-term`. It might be necessary to escape a space (also by placing a `\` in front of it).
 
+### Advanced options to search for combinations of words
+The Elasticsearch query syntax also allows fuzzy matches on a *word* level. This can be used to construct queries in which two words should appear no more than _n_ words apart. For instance,
+>"interest balance"\~5
+
+would find all documents in which the terms "interest" is followed by "balance", separated by no more than 5 words.
+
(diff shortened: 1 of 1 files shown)


frontend/src/assets/manual/en-GB/query.md
```diff
@@ -47,6 +59,7 @@ Illustrating the differences when searching for different combinations of `bank`
 | `bank~1` | 76241 hits (compare with just bank) |
 | `"the bank is"` | 24 hits |
 | `"the bank is" ~1`| 32 hits |
+| `"interest balance"~5 "balance interest"~5` | 285 hits |
```
lukavdplas, line 62: this is almost identical to the row above it, but switches to a different example. Suggestion for clarity:

```suggestion
| `"the bank is"~5 "is the bank"~5` | 285 hits |
```

frontend/src/assets/manual/en-GB/query.md
```diff
@@ -27,6 +27,18 @@ By default the search will combine all terms using `OR`. This means that when yo
 ### Be Careful with Spaces
 Adding or removing a space can change the results of your query. For example search for `+- term` is different than searching for `+-term`. It might be necessary to escape a space (also by placing a `\` in front of it).
 
+### Advanced options to search for combinations of words
```
lukavdplas, line 30: This looks clear! It seems to contradict line 21, though. I think that one needs to be cleared up.

frontend/src/assets/manual/en-GB/query.md
BeritJanssen: I decided to rewrite this and the above examples to _asset_ / _bank_ combinations, as that is what the text on top of the section states.

PR #1730, opened 2025-01-08
Feature/parlamint turkey

Meesch:
This is the corpus definition for the ParlaMint 4.0 Turkey corpus. It includes an updated utils file for Parlamint 4.0 data, which is still a work in progress. I will comment on code that can use improvement in the utils file. For the Turkey corpus I think I am done!


backend/corpora/parliament/utils/parlamint_v4.py
```diff
@@ -98,8 +98,8 @@ def extract_person_data(node):
         'name': name,
         'role': role,
         'gender': gender,
-        'party_ids': party_ids,
-        'party_nodes': party_nodes,
+        'org_ids': org_ids,
+        'org_nodes': org_nodes,  # unsure about this still: might be a bit of a memory leak to store the nodes in the metadata
```
Meesch, line 102: By which I mean: Nodes are pretty big in BeautifulSoup, so to store them all for each person seems not ideal. I can look at this later myself too, just haven't gotten around to it yet. The problem is just that `org_nodes` is currently used to extract the role in the specific corpus definition...

backend/corpora/parliament/utils/parlamint_v4.py
```diff
@@ -0,0 +1,198 @@
...
+    for org in org_data:
+        org_dict[make_id(org['id'])] = org
+    return org_dict
+
+def extract_person_data(node):
+    id = '#' + node['xml:id']
+    surname = node.persName.surname.text.strip()
+    forename = node.persName.forename.text.strip()
+    name = ' '.join([forename, surname])
+    role = node.persName.roleName.text.strip() if node.persName.roleName else None #too simple, needs to be able to have different values and be gotten from the org_node
```
Meesch, line 79: This is currently handled in the specific corpus defition, by a way that requires all org_nodes to be saved to the person, see my other comment.

backend/corpora/parliament/utils/parlamint_v4.py
jgonggrijp: The word is "expensive" or perhaps "inefficient". "Memory leak" means something more specific, which is also more problematic but which is fortunately not what you mean here. A memory leak is always expensive but the opposite is not true.

(I recommend changing the wording because a memory leak is often a serious bug that must be hunted down. No need to put out a red flag here.)

BeritJanssen (commented):
Nice work! Some thoughts:
- how specific is the `ParlamintTurkiye` corpus definition for that country's data. Would it be an idea to make a `Parlamint` subclass that other corpora can inherit from? I can imagine that most field names would be the same
- can you make a `tests/data` directory with some example data? That would make it easier to review, and you'll need it anyway for the unit tests
- without that, I'm not sure this suggestion makes sense, but it seems you get a "role id" from the metadata about the persons, with which you might be able to search the `org` metadata?

backend/corpora/parliament/clarin_parlamint/description/parlamint_turkiye.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Data source and acquisition
+
+The data is scraped from the official web page of the parliament (https://www.tbmm.gov.tr/). The transcripts for this period are published as HTML documents. The data is downloaded using GNU wget, and extracted from the HTML files using Beautiful Soup, and lxml libraries (Python). Except for the default processing built into these libraries (for encoding correction, HTML cleanup), no other preprocessing was applied.
+
+### Corpus-specific metadata
+
+Current version of the corpus contains the sex, the date/place of birth, and when available Wikipedia, Wikidata and Twitter linksfor regular speakers. The corpus also encodes the constituencies of the parliament members. Changes to party affiliations are documented during the time period of the corpus. Otherwise, the start of party affiliation is left unspecified.
```
lukavdplas, line 24: "gender" is more appropriate than "sex" in this context (and also what the field is called in the corpus).

PR #1731, opened 2025-01-09
Gallica: automate restart after failed HTTP requests

BeritJanssen:
There is a throttle on the Gallica API, which caused periodical crashes & necessity for manual restarts. This branch automates those restarts, and it now seems to run smoothly (albeit very slowly) over the remaining Figaro data.


JeltevanBoheemen (approved):
I agree with this solution.

It being slow is not such a bad thing, as long as it keeps going. You could experiment with a dynamic timeout, e.g. wait 5s on the first retry, 10s on the next etc.

PR #1739, opened 2025-01-15
Feature/ner mappings

### LINKED ISSUE #1724: annotated-text mappings may be unnecessary
From what I understand, the current working of named entity annotations is:

- There is a `:ner` field that duplicates the content field, but with annotations in the format `Call me [Ishmael](PER)`
- There are separate `:ner-kw` keyword fields that list named entities per category, which are used for filtering and visualisations.

It occurs to me that the `:ner` field is not used for querying or aggregating, just for viewing a document. (Since the annotations only provide the *type* of entity, rather than its identity, and we have the `:ner-kw` fields as well, I don't expect that querying the `:ner` field will become more relevant in the future.)

So it looks like we could also remove the `annotated-text` mapping for the `:ner` field. As long as the plain-text version is still in the document's `_source`, you have the data to generate the annotations for a document.

<details>
<summary>
Looking at the code, it appears that there is one place where the annotations (rather than the accompanying keyword field) are queried, namely the <a href="https://github.com/CentreForDigitalHumanities/I-analyzer/blob/develop/backend/es/views.py"><code>NamedEntitySearchView</code></a>, but this does not seem unavoidable to me (expand for details).</summary>

This view extracts annotations based on a document ID and is used in the document view to show annotation highlights. It makes a query based on the ID, but also filters the (singular) result further, to only return documents with at least one annotated entity in them.

I *think* this is to represent those document identically to ones without a named entity field. If so: you could check the same thing using the keyword fields. Or check whether there are any annotations after analysing the text. (Since this view is not handling documents in bulk, the extra processing would be negligible.) Or you could revisit whether "there is no data on named entities" and "there are no named entities in this document" are really identical situations.
</details>

Since the *unannotated* text is already indexed, you wouldn't need to index the annotated field for the sake of searching. This also prevents an issue where a field is indexed twice, and so matches in that field are effectively boosted.

Pros:
- Removes the plugin dependency in Elasticsearch
- Consumes less index space
- No issues with how the duplicated field affects relevance scores

(NB: I labelled this "affects-elasticsearch-index", but I think removing the need for an annotated-text mapping won't actually require changes to the index. You just wouldn't use the data anymore.)

BeritJanssen:
Close #1724 : this branch removes the functionality relying on the presence of `annotated_text` mappings. I verified this works on the test server. One thing I'm not too happy with is the way `find_named_entity_fields` function now goes from `corpus.es_index` to the actual (versioned) index name. Will continue with this tomorrow, as I'm sure we already have a proper way to derive this information in place.

Another TO DO: drop the `Annotated Text` plugin in Docker.


lukavdplas:
> One thing I'm not too happy with is the way find_named_entity_fields function now goes from corpus.es_index to the actual (versioned) index name. Will continue with this tomorrow, as I'm sure we already have a proper way to derive this information in place.

Yeah, I think the way you're doing it is fine, but the function you're looking for this this one: https://github.com/CentreForDigitalHumanities/I-analyzer/blob/09317cd2f50f67353effdedd7496ee37604be36b/backend/es/es_alias.py#L54

PR #1741, opened 2025-01-22
GitHub actions: multi platform build for the chip-impaired

BeritJanssen:
This was necessary to make the current setup (with a GitHub actions-built base image) work on my M1 chip. I removed the cache part, since the `pip install` happens every time anyway.


JeltevanBoheemen:
I tested this on my M1 machine, because I hadn't spinned up Docker in a while. With some unrelated problems (lingering volumes mostly), it worked as advertised!

JeltevanBoheemen (approved):

PR #1742, opened 2025-01-22
Store corpus date as integer

### LINKED ISSUE #1733: Min/max data in corpus model: store as integer
The `CorpusConfiguration` model includes the min/max data for the corpus, which is used to show/filter metadata for each corpus in the corpus overview. The page only shows the date as a year though, which is sufficient.

Suggestion:
- Change the field to an integer
- When importing a corpus definition, if min/max is a date, extract the year.
- Change the corpus JSON definition API to use an integer (view/schema/frontend functions /etc)
- Change the corpus form to use number input

lukavdplas:
- Stores the corpus min/max date as a year instead of a date (close #1733)
- Changes the input of the corpus filter (on the home page) and the corpus form to use a number input rather than date input
- Changes the component used in the category dropdown on the corpus form, which gets rid of a weird double border

Internal changes:
- The corpus filter component now uses the `ReactiveFormsModule` for the underlying logic
- The dropdown component now supports `[formControl]` input.


JeltevanBoheemen (approved):
Yeah nice. Was a little bit surprised by the feature creep of this PR, but the description mentions it, so no problem.

PR #1746, opened 2025-01-30
utilise session cleanup for test indices

lukavdplas:
Tiny change to the names used for indices created in unit tests.

The affected indices are now caught by the `test_index_cleanup` fixture, which removes test indices at the start and end of each session. This fixes a problem where tests may fail because the previous session was not completed (and lingering indices are messing up the test).


lukavdplas:
> Nice! Before you merge: can you check if `addcorpus.tests.test_filters.test_date_filter` and `es.tests.test_sync.test_update_index_availability` can lose the direct dependency on `test_index_cleanup`? In principle, they should be using the test client, right?

I checked. Those tests use the `es_client` fixture but that just instantiates an `Elasticsearch` object with the correct settings. Tests or fixtures that create test indices, like these ones, also use the cleanup fixture. So with the current fixture structure, it's not redundant.

PR #1753, opened 2025-02-06
Feature/matomo

lukavdplas:
Adds configuration for analytics using matomo.

- This will import matomo modules if the application environment includes a matomo configuration.
- This includes page tracking, with a custom interceptor to remove query parameters. So we see the visits for a corpus, but not the queries that users enter.


frontend/src/app/routing/matomo.ts
```diff
@@ -0,0 +1,29 @@
+import { inject } from '@angular/core';
+import { NavigationEnd } from '@angular/router';
+import { MatomoModule, MatomoRouterInterceptorFn, MatomoRouterModule, MatomoTracker } from 'ngx-matomo-client';
+
+const removeQueryParams = (url: string) => url.replace(/\?.+$/, '');
+
+const removeQueryParamsInterceptor: MatomoRouterInterceptorFn = (event: NavigationEnd) => {
+    const tracker = inject(MatomoTracker);
+
+    tracker.setReferrerUrl('');
```
JeltevanBoheemen, line 10:
```suggestion
    // hide the referral URL and sanitize the event URL because they contain query parameters
    tracker.setReferrerUrl('');
```

JeltevanBoheemen (changes requested):

Issue #1735, opened 2025-01-15
Upload corpus documentation in form


lukavdplas:
Corpora can include documentation pages, this should be added to the corpus form. Add a separate step for this in the form.

The content of each page is written in markdown, see the [internal documentation](https://github.com/CentreForDigitalHumanities/I-analyzer/blob/develop/documentation/Corpus-documentation.md).

Maybe use a ready-made component to write the markdown files, e.g. codemirror?

The backend API supports uploading the documentation contents, so this just needs a frontend component.

lukavdplas:
Uploading is now possible, but editor needs to be improved; would like to use a WYSIWYG editor.

PR #1755, opened 2025-02-13
Edit corpus documentation

lukavdplas:
Adds a step to add/edit corpus documentation in the corpus form (part of #1735)

The interface is kind of barebones, with just a plain text field. This isn't very user-friendly (even if you know markdown, it's hard to read without syntax highlighting), but at least provides minimum usability. I'll look into adding a "fancier" editor later.

Screenshot:

screenshot of I-analyzer corpus form. Shows a page to add documentation, with tabs for different pages, and a text input field


frontend/src/app/corpus-definitions/corpus-definition.service.ts
```diff
@@ -123,6 +144,47 @@ export class CorpusDefinitionService implements OnDestroy {
...
+            corpus: page.corpusName,
+        };
+        if (page.id && page.content.length) {
+            return this.apiService.updateCorpusDocumentationPage(page.id, data)
+        } else if (page.content.length) {
+            return this.apiService.createCorpusDocumentationPage(data);
+        } else if (page.id) {
+            return this.apiService.deleteCorpusDocumentationPage(page.id);
+        } else {
+            return of(undefined);
```
JeltevanBoheemen, line 184: Was going to recommend a `switch...case` for this but after reading the documentation, let's not.

frontend/src/app/corpus-definitions/form/documentation-form/documentation-form.component.html
```diff
@@ -0,0 +1,32 @@
+<h1 class="title">Add documentation</h1>
+<p class="subtitle">Provide information about your corpus</p>
+
+<p class="block">
+    Add documentation to help users understand what your corpus is, and how they can use
+    it in their research. You can use
+    <a href="https://www.markdownguide.org/">markdown</a> to style the content.
```
JeltevanBoheemen, line 7: Do we allow any markdown flavour like GitHub? May consider adding link to documentation if so.

frontend/src/app/corpus-info/corpus-info.component.ts
```diff
@@ -32,7 +32,7 @@ export class CorpusInfoComponent implements OnInit {
 
     setCorpus(corpus: Corpus) {
         this.corpus = corpus;
-        this.documentation$ = this.apiService.corpusDocumentationPages(corpus).pipe(
+        this.documentation$ = this.apiService.corpusDocumentationPages(corpus.name).pipe(
```
JeltevanBoheemen, line 35: I thought we moved away (as much as possible) from using corpus names? Can we use ID instead? With user uploaded corpora I foresee the unique corpus name restraint to be dropped in the future.

frontend/src/app/models/corpus.ts
```diff
@@ -146,5 +146,20 @@ export interface CorpusDocumentationPage {
...
+    corpus: string,
+    content_template: string,
+};
+
+
+export interface CorpusDocumentationPageSubmitData {
+    type: string,
+    corpus: string,
+    content_template: string,
+};
```
JeltevanBoheemen, line 165: Nitpicky, but the first interface oculd extend the second to avoid some repetition.

frontend/src/app/services/api.service.ts
```diff
@@ -256,6 +257,22 @@ export class ApiService {
...
+
+    public updateCorpusDocumentationPage(pageID: number, data: CorpusDocumentationPageSubmitData) {
+        const url = this.apiRoute(this.corpusApiUrl, `documentation/${pageID}/`);
+        return this.http.put(url, data);
+    }
+
+    public deleteCorpusDocumentationPage(pageID: number) {
+        const url = this.apiRoute(this.corpusApiUrl, `documentation/${pageID}/`);
+        return this.http.delete(url);
+    }
```
JeltevanBoheemen, line 273: We should really build some generic REST endpoints for the API service at some point..

JeltevanBoheemen (approved):

frontend/src/app/services/api.service.ts
lukavdplas: Good point, I'll make an issue.

frontend/src/app/corpus-definitions/form/documentation-form/documentation-form.component.html
lukavdplas: We render documentation with [marked](https://marked.js.org/). They specify what they support, but it's not really a user-friendly guide. We could link to something like like [markdownguide.org](https://www.markdownguide.org/).

JeltevanBoheemen: Yeah, in that case use the existing link.

PR #1758, opened 2025-02-18
Move modules to indexing app

lukavdplas:
This redistributes some of the modules related to indexing and elasticsearch.

The main reason was that we now have an `indexing` app, so it makes sense that functions that are solely concerned with index management (e.g. creating or populating an index) are moved there, while the `es` app is kept for more fundamental elasticsearch business. In that vein, the module to construct an `Elasticsearch` instance is moved to the `es` module.

While moving the modules, I also split up some files that had grown beyond a practical size.

List of changes:

- Move indexing functions and commands, with unit tests, to `indexing` app
- Split `es_index` and `es_alias` modules into smaller modules:
	- `es_index` is split up into three modules: running a `CreateIndexTask`, running a `PopulateIndexTask`, and running an `IndexJob` (i.e. running all tasks).
	- For `es_alias`, functions related to versioning are extracted to a separate module.
- Add one extra unit test :muscle: 
- Move ES client constructor to `es` app
- For versioning-related functions, added typing and clarified the function names and docstrings.

**Note for review:** I did not change any code here, apart from updating imports and extracting some functions into separate modules. Since this code is all quite extensively unit-tested and git diffs for this kind of thing are usually a mess, I would recommend against trying to verify all the code changes. Just check if the list of changes makes sense.


backend/indexing/run_misc_tasks.py
BeritJanssen, line 1: Perhaps this could be called `run_management_tasks` instead (with the idea that some of these tasks would need `manage` rights in Elasticsearch)? Or something else that may be more descriptive than `misc`?

BeritJanssen (approved):

PR #1759, opened 2025-02-20
add create-only option to indexing commands

lukavdplas:
Adds a flag to the `index` and `alias` commands to create the `IndexJob` without running it. 

Right now, this is a bit useless since there is no way to run an existing job (yet), but you could use it to do a "dry run".


backend/indexing/management/commands/alias.py
```diff
@@ -28,7 +28,15 @@ def add_arguments(self, parser):
                 deleted.'''
         )
 
-    def handle(self, corpus, clean=False, **options):
+        parser.add_argument(
+            '--create-only',
+            action='store_true',
```
JeltevanBoheemen, line 33: Always have to look up what `store_true` defaults to. This is correct :)

JeltevanBoheemen (approved):

PR #1760, opened 2025-02-20
Feature/dropdown roles

lukavdplas:
Some minor tweaks to make the dropdown component more accessible:

- The semantic roles of the button and popup were [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) and [menu](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) respectively, but this is more for action menus. If you use a dropdown as a form element, the [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role) and [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) roles are more appropriate.
- Since dropdowns are now form elements, they also need to be explicitly labelled, so I added an extra input `labelledBy` to link the dropdown to the labelling element.
- A dropdown should include some properties like [`aria-controls`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) that refer to the ID of another element ("this controls that"). These IDs were assigned as constants, but this is a problem if you have multiple dropdowns on the same page. This is changed so the IDs are always unique. I think the solution for this is one I hadn't used in I-analyzer before; I got this strategy from [ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/).
- Addressed a slight deviation from the HTML specification in the keyboard controls: if you focus on the dropdown button when it is closed, and press the <kbd>Down arrow</kbd>, the popup should open and focus on the first element. This did not happen; now it does.

Also adds some extra logic to handle focus in the tagging control. E.g. when you select "create new" and the text input appears, you can immediately start typing instead of tabbing to the input first.

### DIFF
--- a/frontend/src/app/shared/dropdown/dropdown.component.ts
+++ b/frontend/src/app/shared/dropdown/dropdown.component.ts
@@ -39,6 +39,7 @@ export class DropdownComponent<T> implements OnChanges, AfterViewInit, OnDestroy
 
     @Input() value: any;
     @Input() disabled: boolean;
+    @Input() ariaLabelledBy: string;
 
     @Output()
     public onChange = new EventEmitter<T>();
--- a/frontend/src/app/shared/dropdown/dropdown.component.ts
+++ b/frontend/src/app/shared/dropdown/dropdown.component.ts
@@ -20,6 +20,8 @@ import { actionIcons } from '../icons';
 import { DropdownService } from './dropdown.service';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 
+let nextID = 0;
+
 @Component({
     selector: 'ia-dropdown',
     templateUrl: './dropdown.component.html',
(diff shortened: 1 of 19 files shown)


frontend/src/app/shared/dropdown/dropdown.component.ts
```diff
@@ -65,6 +74,14 @@ export class DropdownComponent<T> implements OnChanges, AfterViewInit, OnDestroy
         return this.dropdownService.open$.value;
     }
 
+    get triggerID(): string {
+        return `dropdown-trigger-${this.id}`;
```
BeritJanssen, line 78: Could we make the ids more semantic by adding `labelledBy` content instead?

BeritJanssen (approved):
LGTM!

frontend/src/app/shared/dropdown/dropdown.component.ts
lukavdplas: Yes, the button will be named `dropdown-trigger-4` or such, based on the order of construction.

You could allow more descriptive IDs, but since IDs are not visible to the user (unless they inspect the HTML), I don't think that would really have any benefit.

PR #1763, opened 2025-03-06
json schema endpoint + validation

### LINKED ISSUE #1737: API endpoint for corpus JSON schema
Add an API endpoint to retrieve the corpus JSON schema

lukavdplas:
- Adds an API endpoint for the corpus JSON schema (close #1737)
- Validates corpus JSON uploads against the schema. The feedback for invalid JSONs is taken straight from the validation package.

Screenshot:

screenshot of I-analyzer. Shows file upload and an error message detailing issues with an invalid JSON file

### DIFF
--- a/frontend/src/app/corpus-definitions/definition-json-upload/definition-json-upload.component.ts
+++ b/frontend/src/app/corpus-definitions/definition-json-upload/definition-json-upload.component.ts
@@ -4,6 +4,7 @@ import { BehaviorSubject, Observable, Subject, from, of } from 'rxjs';
 import { catchError, filter, map, switchMap, takeUntil, tap, withLatestFrom } from 'rxjs/operators';
 import { actionIcons } from '@shared/icons';
 import { ApiService } from '@services';
+import { ValidationError, Validator } from 'jsonschema';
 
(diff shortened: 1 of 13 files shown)


frontend/src/app/corpus-definitions/definition-json-upload/definition-json-upload.component.ts
```diff
@@ -15,14 +17,18 @@ export class DefinitionJsonUploadComponent implements OnChanges, OnDestroy {
 
     actionIcons = actionIcons;
 
+    schema$ = this.apiService.corpusSchema();
```
JeltevanBoheemen, line 20: I would prefer this to be available in a central location (for example `CorpusService`), but for now this is fine.

JeltevanBoheemen (approved):
Nice. I was also working on this endpoint. I chose to retrieve the schema once in the frontend, and have it available as a subject/behaviorsubject. However, I ran into the issue that it would someties revert to undefined. let's just go with this approach for now.

As discussed in person, the error messages are still quite "techy", would need to improve upon them when releasing this feature to broader audiences.

PR #1764, opened 2025-03-06
use pickle serialiser in celery

### LINKED ISSUE #1757: Use pickle serialiser for celery tasks
See https://docs.celeryq.dev/en/v5.2.7/userguide/calling.html#serializers

Our celery app uses the `json` serialiser, which is the default, but can be cumbersome, because it means task data must be JSON-serialisable. In practice, we often want to pass Django models as the input to a task.

I think in our case, we can just switch to the `pickle` serialiser, which would allow us to use Django models as task input. In general, the advantages of the JSON serialiser are:

- compatability with other languages
- task data will never contain executable code, which is nice for security

I think this is all fine in our case, since we're not sharing our task backend with other programs (or other users).

This is relevant for #1697, which requires making celery tasks to execute various types of `IndexTask`s. In our current codebase, the `download` app also suffers from this issue, where tasks are centred around `Download` objects.

### Implementation

Just add this to the project settings:

```py
CELERY_TASK_SERIALIZER = 'pickle'
CELERY_RESULT_SERIALIZER = 'pickle'
CELERY_ACCEPT_CONTENT = ['json', 'pickle']
```

For bonus points,  the `download.tasks` module can probably be simplified if tasks can get a `Download` object as input.

lukavdplas:
Close #1757

### DIFF
--- a/backend/conftest.py
+++ b/backend/conftest.py
@@ -236,3 +236,11 @@ def json_mock_corpus(db, json_corpus_definition) -> Corpus:
     corpus.configuration.save()
 
     return corpus
+
+@pytest.fixture(scope='session')
+def celery_config():
+    return {
+        'task_serializer': 'pickle',
+        'result_serializer': 'pickle',
+        'accept_content': ['json', 'pickle'],
+    }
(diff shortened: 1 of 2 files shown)


backend/conftest.py
lukavdplas, line 1: I would have expected the test worker to pick up on the global celery config, but apparently it doesn't, so that's why these settings have to repeated in the conftest.

BeritJanssen (approved):

PR #1767, opened 2025-03-13
Feature/update angular

BeritJanssen:
Updates Angular to version 19. Requires Node 22 to run.


BeritJanssen:
@lukavdplas , the failing unit test comes from the `meta-form` component. The Angular update added `standalone: false`, which I believe is correct since this component depends on the `corpus-definitions` module. The problem is that `corpus.definition` is not defined in the `ngOnChanges` of this component. I tried fixing that by providing a `CorpusDefinition` object, but still, the `definition` member is undefined when running the unit test. Help appreciated!

lukavdplas:
Please update the `.nvmrc` file as well!

lukavdplas:
Cherry-picked https://github.com/CentreForDigitalHumanities/I-analyzer/pull/1763/commits/e6459d063f262c983915bd8e65138db4e147b7ab into this branch; that fixes the tests in this branch on my machine.

JeltevanBoheemen:
PrimeNG docs say this about disabling dark mode:
> It is also possible to disable dark mode completely using false or none as the value of the selector.


```providePrimeNG({
    theme: {
        preset: Aura,
        options: {
            darkModeSelector: false || 'none'
        }
    }
})```

BeritJanssen:
Suppressing the `@import` warnings would require upgrading Bulma, which I decided to move to a separate issue (https://github.com/CentreForDigitalHumanities/I-analyzer/issues/1773). Should also investigate a better way to deal with variables, but since the Bulma upgrade might affect the use of color variables, perhaps best to defer this to this issue. Perhaps we can also use a color mixer function from Bulma to compute the different gradient values of the `primary` color.

frontend/src/app/services/api.service.spec.ts
lukavdplas, line 1: The field selection switches between showing and not showing the value:

screenshot of field selector with label "all fields"
screenshot of field selector with no label

lukavdplas:
Tiny hill: some elements now have a colour that is *slightly* off from I-analyzer blue. Here you can see that the "search" and "page 1" buttons are slightly different colours of blue (`#303f9f` and `#485fc7` respectively)

screenshot of I-analyzer search interface

lukavdplas:
Not a fan of how, in this case, the disabled filters stand out more than the available ones:

screenshot of I-analyzer showing several multiple-choice filters

lukavdplas:
In the navbar links, there should be some extra space between the icons and the link text. Compare the old version:

screenshot of I-analyzer navbar

and the new one:

screenshot of I-analyzer navbar

lukavdplas:
Something's wrong with the spacing in the footer here:

screenshot of document popup in I-analyzer

lukavdplas:
Hm, can't make a screenshot of this one. But if you open the P&P NL corpus on the test server, go to search results and open the "sort by" dropdown, there is some inconsistent indentation.

lukavdplas:
z-index triumphs again:

screenshot of I-analyzer search interface. Shows user menu is partially obscured by search section

lukavdplas:
Update:

- Updated branch from develop
- Fixed all issues mentioned above, including #1773

PR #1776, opened 2025-04-07
delpher documentation

lukavdplas:
Part of https://github.com/CentreForDigitalHumanities/I-analyzer/issues/991

Adds an info page for the public Delpher corpus. Also includes a (Jill-approved) explanation about accessing the full dataset.


JeltevanBoheemen (approved):
Jill-approved === everyone approved

PR #1782, opened 2025-04-15
manage/schedule index jobs from CLI

lukavdplas:
This adds some functionality to the backend CLI that is currently only available on the admin site:

- View index jobs with `indexjob list` and `indexjob show `
- Start created jobs with `indexjob start`
- When running index jobs from the command line, use `--async` to run them through celery.

### DIFF
--- a/backend/indexing/celery_utils.py
+++ b/backend/indexing/celery_utils.py
@@ -1,0 +1,21 @@
+import warnings
+
+from django.conf import settings
+from ianalyzer.celery import app
+
+
+def worker_is_active() -> bool:
+    '''
+    Check whether the app can connect to an active worker.
+    '''
+    inspector = app.control.inspect()
+    try:
+        workers = inspector.active()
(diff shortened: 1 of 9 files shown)


backend/indexing/celery_utils.py
JeltevanBoheemen, line 1: Appreciate this utils file, but think it should be in a more general module (Celery relates to more functionality than indexing)

backend/indexing/management/commands/indexjob.py
```diff
@@ -0,0 +1,117 @@
...
+            delimiter='\t',
+        )
+        writer.writeheader()
+        for job in jobs:
+            writer.writerow({
+                'id': job.id,
+                'corpus': job.corpus,
+                'created': job.created,
+                'status': job.status()
+            })
```
JeltevanBoheemen, line 84: Neat

JeltevanBoheemen (approved):

PR #1783, opened 2025-04-17
Feature/gallica images

Meesch:
This branch adds a request_media method to the Gallica corpus definition. Tested locally and works like a charm. Oops I also merged develop into this branch so if you merge Caricature first it should just contain a single commit.


ar-jan:
Or you could cherry-pick the image path commit and add it to the Caricature PR for a probably cleaner history (if I understand correctly).

Meesch:
> Or you could cherry-pick the image path commit and add it to the Caricature PR for a probably cleaner history (if I understand correctly).

Ah that is indeed better! Thank you for the tip! Juliwan-Kenobi will be pleased.

Meesch:
Okay disregard I have added this commit to the feature/caricature branch PR and I have approved it, @BeritJanssen you can take a look there and see if you agree!

PR #1785, opened 2025-04-24
change username

lukavdplas:
Small feature: adds the option to change your username in the settings menu.

### DIFF
--- a/frontend/src/app/settings/change-username/change-username.component.ts
+++ b/frontend/src/app/settings/change-username/change-username.component.ts
@@ -23,6 +23,7 @@ export class ChangeUsernameComponent {
 
     invalidFeedback: string;
     successFeedback: string;
+    errorFeedback: string;
 
     constructor(
         private authService: AuthService,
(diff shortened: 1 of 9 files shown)


frontend/src/app/settings/change-username/change-username.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    submitName() {
+        this.invalidFeedback = undefined;
+        this.successFeedback = undefined;
+        this.errorFeedback = undefined;
+
+        this.authService.updateSettings(this.form.value).subscribe({
+            next: this.onSuccess.bind(this),
+            error: this.onRequestFail.bind(this),
```
JeltevanBoheemen, line 44: Won't `next: this.onSuccess()` work?

JeltevanBoheemen (approved):

frontend/src/app/settings/change-username/change-username.component.ts
lukavdplas: That would be `next: this.onSuccess` or `next: () => this.onSucces()`. (You provide the function, not the outcome.)

The first one (without bind) would not work because `onSuccess` is a bound function (i.e. it references `this`). You could do the second. I'm 50/50 on which is preferable.

PR #1787, opened 2025-04-24
Bugfix/corpus min max year

### LINKED ISSUE #1786: Error when creating corpus through form
### What went wrong?

When you create a new corpus in the corpus form, it won't be created properly; the `CorpusConfiguration` is not created. After that, the definitions overview will fail to load due to a server error.

The error is created by https://github.com/CentreForDigitalHumanities/I-analyzer/pull/1742 ; the `min_year` / `max_year` properties of the corpus configuration are non-nullable fields but they are not provided on creation.

Corpora should be created without errors.

### Screenshot

_No response_

- [ ] https://ianalyzer.hum.uu.nl
- [x] a local server

### Version

5.18.0

### Steps to reproduce

- Log in as an admin user
- Go to the "corpus definitions" menu
- Create a new corpus, enter a title and save.
- Go back to the definitions overview.

lukavdplas:
Fixes some errors surrounding the `min_year`/`max_year` fields for corpora. Close #1786 

- Corrected the definition for the admin form (which listed outdated field names)
- Made the `min_year`/`max_year` fields nullable in the database
- Added validation that the min/max year must be specified for publication.


lukavdplas:
@JeltevanBoheemen re. the nullable fields: I couldn't find any reason why these fields can't be nullable, so I just went with that. It makes more sense than filling in random dates.

JeltevanBoheemen (commented):
Can you walk me through the decision to make these fields nullable? Just going by this PR, two things seem happen:
- Default values are set for `min_year` and max_year`
- Validation checks for non-null

For what cases do we need `null` then?

JeltevanBoheemen:
I understand the validation logic btw, and that it does not conflict with having a nullable field (e.g. "I'm gonna decide on these later"). But a default option seems fine in that case.

lukavdplas:
The `null` value is so the values can be left blank when you create the corpus. A default value would also solve that, but there isn't really a neutral or typical value for the time range. It makes more sense to me that the fields are blank when you first see the form. 

I guess I would say: a default option make sense if some (or preferably most) users can look at the value and move on to the next question. So it saves time/effort/complexity in the form. But in this case, pretty much any corpus is going to require that the creator fills in the specific time period, a default period does not really save time.

JeltevanBoheemen:
I agree with that reasoning. However, these are **also** default values. I would be in favour of removing those, they are kind of an uneducated guess anyway.
```
min_year = models.IntegerField(
        help_text='earliest year for the data in the corpus',
        default=DEFAULT_MIN_YEAR,
        null=True,
        blank=True,
    )
max_year = models.IntegerField(
        help_text='latest year for the data in the corpus',
        default=default_max_year,
        null=True,
        blank=True,
    )
```

lukavdplas:
Lol, I completely missed those. Yes, those should be removed.

PR #1789, opened 2025-04-30
Corpus ownership

### LINKED ISSUE #977: Add curator role
Add a 'curator' user role, who will have the right to create and maintain corpora but is not an admin.

lukavdplas:
Changes the permissions surrounding corpora:
- Whether a user can create corpora is now a separate field from whether they are an admin. 
- The database now tracks the ownership of corpora. (When you create a corpus, you automatically become the owner.) You can only edit corpora that you own.
- Owning a corpus also gives you a free pass to search it (if it is active), even if you are not added to any relevant group.

Close #977 

Also tried to make the corpus permissions a bit more consistent. Some permission-checks were implemented in `addcorpus.permissions` and some in `users.models` -> now everything is in `addcorpus.permissions`. Also, there is a single function that defines the condition for search rights, idem for editing rights.


lukavdplas:
By @JeltevanBoheemen 's suggestion: changed corpus ownership field to many-to-one.

JeltevanBoheemen (approved):

Issue #1795, opened 2025-05-14
Corpus form: CSV upload flow


lukavdplas:
The upload step in the  source data allows the user to upload CSV data and reset their uploaded file. In practice, users will probably want to replace the data file. This can be to edit the data model (e.g. add fields, change data types), or without changing the data model (e.g. add more documents, correct content).

The current situation needs improvement:
- When you upload a new file, the fields for the corpus are generated based on the file. So if you want to update the file, you lose all field settings.
- To upload a new file, you must first remove the old one, and this can't be reversed.

The attached file describes a new workflow in detail that @JeltevanBoheemen and I agreed on.

[Corpus form - csv upload.md](https://github.com/user-attachments/files/20205942/Corpus.form.-.csv.upload.md)

lukavdplas:
This would also resolve #1780

PR #1797, opened 2025-05-14
Feature/steps logic

lukavdplas:
Builds on #1792. Some improvements to the workflow in the corpus form:

- Moved the documentation form to step 1 (corpus information), instead of making it a separate step.
- Added success / error messages and loading spinners to save buttons in steps 1 (corpus info) and 3 (fields). Left step 2 as-is because that one still needs more work (#1795)
- Changed the logic for disabling steps; this is now done by checking the content after every save. The exact requirements will probably still need some tweaking.
- Added a button on the bottom of the page to move to the next step.


JeltevanBoheemen (approved):
Good changes. I find the UI flow to get to the next step a bit unclear still. Filling it in I'm left wondering which information I'm supposed to change and which are optional. This is beyond the scope of this PR though. Good work.

PR #1798, opened 2025-05-14
search UI improvements

lukavdplas:
These are a few UI and accessibility improvements to the search interface:

- The results count is now a live region, so when you update the query, assistive technologies will announce that the results have been updated.
- Results count now use the `number` pipe so you can see the difference between 100,000 and 1,000,000 results.
- Added aria labels to filter controls
- Better semantic roles: the filter menu is now a `<form>` and each panel a `<fieldset>`. 
- Filters now have a separate question mark icon that spawns the tooltip (rather than having the tooltip on the title). As I discussed wilth @JeltevanBoheemen , I think we should do this more often. If the element is not interactive, the user may never hover over it and not realise there is extra clarification available. 
- I always have trouble understanding the state of the checkbox control for boolean fields. Changed to a dropdown with options  "any value", "true", "false".


frontend/src/app/filter/boolean-filter/boolean-filter.component.html
```diff
@@ -1,4 +1,11 @@
 <div class="bool" *ngIf="filter">
-    <p-checkbox [ngModel]="data" binary="true" (onChange)="update($event.checked)"></p-checkbox>
-    {{data | json | titlecase }}
+    <ia-dropdown [value]="data" (onChange)="update($event)"
+        labelledBy="legend-filter-{{filter.displayName | slugify}}">
+        <span iaDropdownLabel>{{data !== undefined ? (data | json) : 'any value'}}</span>
+        <div iaDropdownMenu>
+            <a iaDropdownItem>any value</a>
```
JeltevanBoheemen, line 6: Not big on this label. It obfuscates the fact that the filter is a boolean. But also hard pressed to find a better alternative.

`True or False`? (might get into arguments with logicians with this kind of wording..)

JeltevanBoheemen (commented):
Really nice improvements. The number formatting alone is quite the upgrade :)

Concerning the questionmark-hover-buttons-not-buttons:
- Love 'em
- Tooltip behaviour in the application is now mixed; some elements have a tooltip, have the dedicated icon. Not something I mind terribly, but to keep in mind. 
- Tooltip implementation will also be harder going forward, requiring positioning of this extra UI element. Again, I think the pro outweighs the cons.

frontend/src/app/filter/boolean-filter/boolean-filter.component.html
lukavdplas: Hm, true, the label is not great.

Re. "true or false": the default state of the filter allows `{ true, false, null }`, not `{ true, false }`. So logic pedantry aside, "true or false" does not really reflect the value, if the field is empty for some documents. Though to be fair, the label "any value" also does not really reflect that this includes "no value".

(Also: yes, this makes the distinction between booleans and keyword fields less clear, but for the user, that distinction is perhaps not that important?)

Issue #1697, opened 2024-11-08
IndexJob lifecycle


lukavdplas:
Part of #1007

After #1696, there should be an `IndexJob` that records indexing tasks, but is essentially just a log.

Refactor the indexing functionality so it is centred around the lifecycle of an `IndexJob`. More concretely, allow administrators to create jobs in the admin site that will actually be run.

- You can create a new `IndexJob` using the admin site. This just creates the object with status "pending".
- Using an admin action, you can start a pending job. I-analyzer will use celery to execute the job asynchronously. Status is updated to "in progress".
- You can interrupt a job using a Django admin action. This kills the task in celery and sets the status to "aborted".
- When the celery task completes or fails, the status is updated to "complete" or "error".

### CLI

It's possible to keep the `index` and `alias` commands function as they do now. Under the hood, those commands would need to:
- Create an `IndexJob` that describes the action (this should already be implemented in #1696)
- Immediately start the job, and run the celery task synchronously.

Alternatively, you could rework the command line interface to be parallel to the admin actions. So you use `index` and `alias` to create the IndexJob without running it. Then you could add commands to view, start, and stop index jobs. (But you could also create a job in the command line and then start it from the admin site, or vice versa.)

PR #1799, opened 2025-05-15
interrupt index jobs

lukavdplas:
Allows you to stop index jobs that are run asynchronously, either via the admin or the command line.

Completes #1697


joao-krauchenco-mottu (approved):

backend/indexing/models.py
```diff
@@ -127,6 +127,11 @@ class Meta:
         default=TaskStatus.CREATED,
         help_text='execution status of this task',
     )
+    celery_task_id = models.CharField(
+        max_length=128,
+        null=True,
+        blank=True,
```
JeltevanBoheemen, line 133: Why nullable?

backend/indexing/run_job.py
```diff
@@ -37,14 +40,26 @@
 def run_task(task: IndexTask) -> None:
     '''Run an IndexTask'''
     task_id = f'{task.__class__.__name__} #{task.pk}' # e.g. "CreateIndexTask #1"
+
+    try:
+        raise_if_aborted(task)
```
JeltevanBoheemen, line 45: This seems convoluted. Will the `raise` not occur in the natural flow of indexing (when the task is aborted)? Now it seems like i you forget to ask specifically for a `raise` check, the program fails silently.

JeltevanBoheemen (commented):
Nice managment interface for stopping.
I have some doubts about the `raise_if_aborted` mechanism. This goed against my (maybe unjustly so) feeling that a `raise` should occur whenever something fails, and you should not need to ask explicitly "we still okay?"

Ofcourse open to explanation as to why this should work like this.

PR #1800, opened 2025-05-20
Hotfix/5.18.1

JeltevanBoheemen:
Downloads directory is growing out of bounds with incoming requests.

This hotfix implements a signal that removes downloaded files from when the model is deleted.

Proposed query to clean up the csv directory:
`to_delete = Download.objects.exclude(completed__isnull=True).filter(user__isnull=True)`
This removes all completed anonymous downloads.


BeritJanssen (approved):
👍

PR #1807, opened 2025-05-27
Feature/download rate limit

### LINKED ISSUE #1804: Download rate limit
Recent overload of server through mass download requests stresses the need for a rate limited download API.

Implement a sensible rate limit that respects scraping and manual use, while preventing DDoS like behaviour.
Limit of 5 requests/minute/user is a good starting point. Implementation can be copied from recent rate limiting to login and registration endpoints.

Closing the download API to unauthenticated users does not work:
- Users with an account can still spam the hell out of the API
- Data access becomes restricted while we value the open data nature of public corpora.

JeltevanBoheemen:
Closes #1804

Implements a rate limit for the direct downloads perfomed by anonymous users.

### DIFF
--- a/backend/download/tests/test_download_views.py
+++ b/backend/download/tests/test_download_views.py
@@ -12,6 +12,7 @@ from addcorpus.models import Corpus
 from visualization import query
 from es.search import hits
 from tag.models import Tag, TaggedDocument
+from django.core.cache import cache
 
 
 def test_direct_download_view(admin_client, mock_corpus, index_mock_corpus, csv_directory):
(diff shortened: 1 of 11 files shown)


backend/download/tests/test_download_views.py
```diff
@@ -231,6 +232,7 @@ def test_download_with_tag(db, admin_client, small_mock_corpus, index_small_mock
 
 
 def test_unauthenticated_download(db, client, basic_mock_corpus, basic_corpus_public, index_basic_mock_corpus):
+    cache.clear()  # clear cache to avoid throttle limit
```
lukavdplas, line 235: Since this is repeated several times, perhaps you could create a `clear_cache` fixture for these tests?

I think it would be fine to add an autouse to the fixture as well: there should never be side effects from the cache being preserved between unit tests.

lukavdplas (approved):

backend/download/tests/test_download_views.py
JeltevanBoheemen: Yeah good point. I'll use an auto-use fixture so we don't need to repeat this code for any future tests.

PR #1813, opened 2025-06-10
Ingest European Parliament data from API

BeritJanssen:
This branch adds a way to add resources of the European Parliament from [the OpenData portal](https://data.europarl.europa.eu/en/developer-corner/opendata-api), so far only available for the most recent term (10). Depends on [these updates for ianalyzer-readers](https://github.com/CentreForDigitalHumanities/ianalyzer-readers/pull/34), requirements in this PR should be adjusted to the next release of ianalyzer-readers with these fixes.

I also went ahead and removed the  `addcorpus.python_corpora.extract` module (now unneeded, since we have all extractors in the `ianalyzer_readers` module). Let me know if this change should be reverted due to dependencies I cannot oversee.

The cleanup meant import statements changed in many files. The relevant (to be reviewed) changes are in `corpora.parliament`, `corpora.jewishmigration` and `corpora.gallica`.


lukavdplas:
> I also went ahead and removed the addcorpus.python_corpora.extract module (now unneeded, since we have all extractors in the ianalyzer_readers module).

Sure! I think I had left it like this to avoid changing all the existing imports, but this is cleaner.

lukavdplas (approved):
Looks good!

PR #1814, opened 2025-06-11
Add `word_model_path` for Dutchnewspapers corpora

BeritJanssen:
This branch prepares adding the models of 19th century data (older is not feasible due to mixed language).


lukavdplas:
Could you also add user documentation for these models?

JeltevanBoheemen (approved):

PR #1815, opened 2025-06-11
Update Netherlands Recent corpus definition to read from Parlamint v4 data

BeritJanssen:
The current branch uses the `parlamint_v4.py` utils file from https://github.com/CentreForDigitalHumanities/I-analyzer/pull/1730. I changed it in the following ways:
- added a function to read people and party metadata from the external files (included in the test data)
- added various functions for named entity fields, and parsers to enrich with named entity annotations
- added functions to flatten the annotated format of the speeches
- removed functions that were copied without change from `parliament.utils.parlamint.py` utils


backend/corpora/parliament/utils/parlamint_v4.py
```diff
@@ -0,0 +1,322 @@
...
+from corpora.parliament.utils.parlamint import (
+    metadata_attribute_transform_func,
+    person_attribute_extractor,
+)
+
+
+"""
+This file was created as an updated utils file for the ParlaMint dataset, version 4.0. The previous utils file
+is based on version 2.0.
+"""
```
lukavdplas, line 22: Small note: module level docstrings should be the first line of the file. If you put it after the imports, it's not registered by `help`, code editors, etc.

I think by "the previous utils file", you're referring to `parliament.py`? This reference is ambiguous. Also, someone not familiar with the code history would not know which file came first chronologically.

If I were looking over a directory with `parliament.py` and `parliament_v4.py`, I would probably assume that `parliament.py` covers the most recent version and should be the default for new corpora, and the `v4` module is for compatibility with some older version.

backend/corpora/parliament/utils/parlamint_v4.py
```diff
@@ -0,0 +1,322 @@
...
+        if (start_date and end_date and start_date <= date <= end_date) or \
+        (start_date and start_date <= date) or \
+        (end_date and end_date >= date):
+            return True
+        else:
+            return False
+    else:
+        return False
+
+def transform_current_party_id(data):
```
lukavdplas, line 181: This could do with a docstring. What is `data`? What is is transformed into?

Based on the name, I assumed this function transformed the current party ID into something else, but based on the code, it looks like this function retrieves the current party ID from other data? If so, the name is a bit misleading.

lukavdplas (approved):
The code looks fine. I added some requests for clarity, but other than that, feel free to merge.

backend/addcorpus/es_mappings.py
```diff
@@ -100,5 +100,5 @@ def geo_mapping():
     return {'type': 'geo_point'}
 
 
-def annotated_text_mapping():
-    return {'type': 'annotated_text'}
+def ner_mapping():
+    return {'type': 'text', 'index': False}
```
Meesch: What was the issue with the 'annotated_text' type?

lukavdplas: See https://github.com/CentreForDigitalHumanities/I-analyzer/issues/1724 

TLDR: it wasn't necessary because we already have the unannotated text for full-text search, and keyword fields for finding entities.

PR #1836, opened 2025-07-10
Export tags and document link in .csv

### LINKED ISSUE #1195: Support extra csv columns for search results download
Add the following (optional) columns to the search results CSV:

- The URL of the document in the interface  (e.g. https://ianalyzer.hum.uu.nl/document/great-corpus/awesome-document )
- The user's assigned tags

BeritJanssen:
Close #1195, also updated the download manual, which still described the situation prior to the download module.

### DIFF
--- a/backend/download/conftest.py
+++ b/backend/download/conftest.py
@@ -4,7 +4,7 @@ import os
 from addcorpus.models import Corpus
 from corpora_test.mixed_language.multilingual_mock_corpus import SPECS as ML_MOCK_CORPUS_SPECS
 from download import tasks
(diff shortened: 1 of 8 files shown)


lukavdplas:
I was looking at the [list of forbidden field names](https://github.com/CentreForDigitalHumanities/I-analyzer/blob/759b8f07b4bd4e71387138598546d28e5e100625/backend/addcorpus/constants.py#L40); the docstring correctly points out that field names cannot overlap with any fixed column names, so any columns here must be added to the list.

PR #1840, opened 2025-07-21
19th Century US Newspapers

### LINKED ISSUE #1744: Corpus: 19th Century US Newspapers
- [ ] Request ITS share
- [ ] Test data
- [ ] Corpus definition
- [ ] Indexing
- [ ] Deployment

bbonf:
Most of the `GaleCorpus` base class has been copied from `periodicals.py`
I added a `process_scans` management command for, well, processing scans.
I'm not sure whether falling back to on-the-fly processing is a good idea or not.


backend/corpora/gale/gale.py
```diff
@@ -0,0 +1,309 @@
...
+        extractor=extract.Metadata("title"),
+        csv_core=True,
+        visualizations=["resultscount", "termfrequency"],
+    )
+    content = FieldDefinition(
+        name="content",
+        display_name="Content",
+        display_type="text_content",
+        description="Text content.",
+        es_mapping=main_content_mapping(True, True, True, "en"),
```
lukavdplas, line 136: Are all Gale corpora in English? Otherwise this should be dynamic too.

backend/corpora/gale/gale.py
```diff
@@ -0,0 +1,309 @@
...
+        description="Source issue number.",
+        es_mapping=keyword_mapping(),
+        results_overview=False,
+        extractor=extract.Metadata("issue_id"),
+        csv_core=False,
+    )
+    periodical = FieldDefinition(
+        name="periodical",
+        display_name="Periodical name",
+        histogram=True,
```
lukavdplas, line 120: You can leave out `histogram` here, it's not an argument for `FieldDefinition`. (The Periodicals corpus includes it; I guess it was an argument at some point.)

backend/corpora/gale/gale.py
```diff
@@ -0,0 +1,309 @@
...
+            yield xml_file, meta
+
+    def process_scan(self, src):
+        dst = src
+        base, ext = os.path.splitext(src)
+        if ext.lower() == '.tif':
+            dst = base + '.png'
+            if not os.path.isfile(dst):
+                PIL.Image.open(src).save(dst)
+        return dst
```
lukavdplas, line 72: As I understand, a single TIFF image can contain multiple images (e.g. a multi-page article). Just to double check that that is not the case here.

backend/corpora/news_us_19/test_news_us_19.py
```diff
@@ -0,0 +1,13 @@
+import os
+from corpora.utils_test import corpus_from_api
+
+here = os.path.abspath(os.path.dirname(__file__))
+
+def test_times(settings, db, admin_client):
```
lukavdplas, line 6:
```suggestion
def test_news_us_19(settings, db, admin_client):
```

backend/indexing/management/commands/process_scans.py
```diff
@@ -0,0 +1,29 @@
...
+        parser.add_argument(
+            'corpus',
+        )
+
+    def handle(self, corpus, **kwargs):
+        corpus_definition = load_corpus_definition(corpus)
+
+        client = elasticsearch(corpus)
+        index = get_index(corpus)
+        results = client.search(index=index)
```
lukavdplas, line 23: This method will fetch 20 documents by default, which is not what you want; you can provide a `size` but it's capped at 10.000. You can use [scan](https://elasticsearch-py.readthedocs.io/en/v7.15.2/helpers.html#scan) to get all documents in the index.

lukavdplas (approved):
Nice, this looks really good. I added some minor comments; some are just about the Periodicals corpus, which is quite old. There is a bug in the process_scans command; all other comments are optional.

Re. on-the-fly processing: I'm not sure either. It would be an issue if processing images would result in request timeouts, but I assume it doesn't take that long?

backend/corpora/news_us_19/description/19thCenturyUSNewspapers.md
lukavdplas, line 1: Ah, I forgot this, but: this description could use some extra context about where the dataset is coming from.

As an example of what I mean, see the "read more" and "availablity" sections of these two corpora:

- https://github.com/CentreForDigitalHumanities/I-analyzer/blob/develop/backend/corpora/ecco/description/ecco.md
- https://github.com/CentreForDigitalHumanities/I-analyzer/blob/develop/backend/corpora/guardianobserver/description/guardianobserver.md

PR #1844, opened 2025-07-25
improve field auto import

lukavdplas:
Small improvement to the field import from CSV files. This now distinguishes between `text_content` and `text_metadata` fields. (Distinction is based on length + line breaks.)

Corpora must have at least one content field to pass validation, so this means a typical corpus is actually valid after the CSV upload.

Also, the labels for datatypes in the `SampleUploadComponent` now match the ones in the field form.

### DIFF
--- a/backend/addcorpus/json_corpora/csv_field_info.py
+++ b/backend/addcorpus/json_corpora/csv_field_info.py
@@ -1,0 +1,48 @@
+import datetime
+import os
+from typing import Dict, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+
+def get_csv_info(path: Union[str, os.PathLike], **kwargs) -> Dict:
+    df = pd.read_csv(path, **kwargs)
+    info = {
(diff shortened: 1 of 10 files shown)


backend/addcorpus/json_corpora/csv_field_info.py
```diff
@@ -0,0 +1,58 @@
...
+def is_date(input: str) -> Optional[datetime.datetime]:
+    try:
+        datetime.datetime.strptime(input, '%Y-%m-%d')
+        return True
+    except (ValueError, TypeError):
+        return False
+
+
+def is_long_text(value: Any) -> bool:
+    return isinstance(value, str) and (len(value) > 100 or '\n' in value)
```
JeltevanBoheemen, line 54: Feels arbitrary, but also sensible - will probably work in like 95% of cases? Good thought.

JeltevanBoheemen (approved):

PR #1870, opened 2025-08-20
Avoid overlap between field names and HTML IDs

lukavdplas:
Implements a less clumsy fix for situations where field names overlap with HTML element IDs. This was done by marking some names as "forbidden"; this uses a more reliable fix, by naming the field element `field-*`


jgonggrijp:
Question as a curious bystander: why is it a problem if the field name and the element ID coincide?

PR #1874, opened 2025-08-21
fix date picker  onChange

### LINKED ISSUE #1871: Text entry in date filter does not trigger submit
### What went wrong?

When you use the date filter and type a date directly into the text field, then hit <kbd>Enter</kbd>, nothing happens.

The expected behaviour would be that the enter key would submit the form. In this case, the query should be updated and results should reload.

### Screenshot

_No response_

- [x] https://ianalyzer.hum.uu.nl
- [ ] a local server

### Version

5.21.0

### Steps to reproduce

- Go to a corpus with a date filter (I used P&P Netherlands)
- Use the text entry field in the date filter to type a new date
- Hit Enter

lukavdplas:
close #1871

### DIFF
--- a/frontend/src/app/corpus-selection/corpus-filter/date-picker/date-picker.component.ts
+++ b/frontend/src/app/corpus-selection/corpus-filter/date-picker/date-picker.component.ts
@@ -1,6 +1,7 @@
-import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { Component, DestroyRef, EventEmitter, Input, Output } from '@angular/core';
 import * as _ from 'lodash';
-import { parse } from 'date-fns';
+import { FormControl } from '@angular/forms';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
 @Component({
     selector: 'ia-date-picker',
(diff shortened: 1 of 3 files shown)


frontend/src/app/corpus-selection/corpus-filter/date-picker/date-picker.component.ts
```diff
@@ -12,30 +13,30 @@ export class DatePickerComponent {
...
-            const dateFormat = this.unit === 'year' ? 'YYYY' : 'DD-MM-YYYY';
-            return parse(value, dateFormat, null, null);
-        } else {
-            return value;
-        }
+    ngOnInit() {
+        this.control.setValue(this.value || this.default);
+        this.control.valueChanges.pipe(
+            takeUntilDestroyed(this.destroyRef)
+        ).subscribe(() => this.onValueChange());
```
jgonggrijp, line 33: I was curious about this PR because you asked me about the `InputEvent`. I find it interesting how you found a more "Angular-native" solution.

Do I understand correctly that you are using the `takeUntilDestroyed` operator to ensure that the subscription closes when the lifespan of the component instance ends—in other words, to prevent a memory leak?

lukavdplas: Yes, `takeUntilDestroyed` is to close the subscription when the component is destroyed. Actually, it may be superfluous in this case? (Textbook use is when you subscribe to observables that may outlive the component instance, e.g. from a service.)

The `InputEvent` question was not related to this PR, actually. (The context was that I wanted to signal something to a non-angular JS library. Specifically [binding Vega signals to external elements](https://vega.github.io/vega/docs/signals/#binding-directly-to-external-elements).)

jgonggrijp: I don't know whether `takeUntilDestroyed` is needed for component-internal observables; that depends entirely on whether the garbage collector in the browser is able to clean up circular references. It's a circular reference because the component holds a reference to the observable, which holds a reference to the subscribing closure, which closes over `this`, which is the component.

In Backbone, a best practice for views (=components) is to call `.remove()` at the end of their lifetime. This cuts off all event listeners, exactly in order to prevent memory leaks by circular references. That makes me think that the circularity might actually be a problem here.

In any case, I found it interesting to learn about `takeUntilDestroyed`.

PR #1881, opened 2025-09-02
Feature/downloads cleanup

lukavdplas:
Adds a command to delete old downloads. Related to https://github.com/CentreForDigitalHumanities/I-analyzer/issues/1805, but does not include periodic scheduling.


JeltevanBoheemen (approved):
noice.

The check is a good idea (always abit scary to delete files). Scheduling could be something like this:
- send an email to admin account rounding up all files to be deleted
- if no action taken, delete them X days later
- have some contingency button to stop the delete

PR #1919, opened 2025-11-05
Feature/corpus form nav

### LINKED ISSUE #1817: Corpus form navigation
The navigation to the corpus form in the site was chosen somewhat arbitrarily and is not very intuitive. Also, the page is called "corpus definitions" in titles and navigation elements, which does not really match the user experience.

(Both of these made more sense when I thought of this form as a kind of admin/moderator interface for corpora. But at present, it looks like all "our" corpora will continue to be created in Python, while the form is used for corpora managed by a single, non-staff user.)

Suggested:

- Rename the page to "my corpora"
- Make this an item in the navbar

What I dislike here is that you would get "corpora" and "my corpora" next to each other. But "my corpora" is not a subset of the "corpora" page; it leads to the edit form, so it has a very different function.

lukavdplas:
close #1817 

Changes to the site navigation to reach the corpus form.

- Corpus form is now called "custom corpora"
- Custom corpora can be reached through the user menu.
- The list of corpora uses cards instead of table, which makes more sense if you only have one or two corpora.
- Added links to edit a corpus from the search interface / corpus overview (if permitted).

Screenshots:

Custom corpora overview:

image

Search interface for a corpus that I own:

image


frontend/src/app/corpus-definitions/form/corpus-form/corpus-form.component.html
```diff
@@ -1,14 +1,35 @@
...
+                <li><a [routerLink]="['../..']">Custom corpora</a></li>
                 <li class="is-active" aria-current="page"><a>{{(corpus$ | async).definition?.meta.title ||
                         '...'}}: edit</a></li>
             </ul>
         </nav>
 
+        <div class="columns">
+            @let corpus = corpus$ | async;
+            <div class="column">
+                <h1 class="title">Edit "{{corpus?.definition?.meta?.title || '...'}}"</h1>
```
JeltevanBoheemen: [Elvis is alive!](https://en.wikipedia.org/wiki/Elvis_Presley_death_conspiracy_theories)

JeltevanBoheemen (approved):

PR #1923, opened 2025-11-07
Use dynamic name when applicable

lukavdplas:
Changes a lot of instances where templates etc. hard-coded "I-analyzer" when they should rely on the `SITE_NAME`/ `appName` configuration.

Does not change the user manual, which does not support dynamic values (cf. https://github.com/CentreForDigitalHumanities/I-analyzer/discussions/1847)


lukavdplas:
Ah, I see what you mean, though at least in our case, I'm hoping we will not make *another* name change, haha.

Issue #1925, opened 2025-11-12
Name change


lukavdplas:
### Discussed in https://github.com/CentreForDigitalHumanities/I-analyzer/discussions/1820

<div type='discussions-op-text'>

<sup>Originally posted by **lukavdplas** June 18, 2025</sup>
We are planning to change the name of I-analyzer to Textcavator. This will probably happen ~~somewhere this autumn~~ this december.

The intention is to change the name everywhere at once to maintain consistency. I'm making this discussion to track everything that needs to happen.

### Codebase
- [x] update all mentions of "i-analyzer" in templates, documentation, etc. -> #1926 
- [x] update development settings in `settings.py` and `environment.ts` -> #1926 
- [x] add some explanation for users in the frontend -> #1936 
- [x] change i-analyzer repository name
- [ ] rename backend app
- [ ] change app prefix in angular
- [ ] rename docker images

### Server
- [x] update production settings in `settings.py` and `environment.ts`
- [x] update server configuration (new repository URL)
- [x] deploy on new domain name
- [x] update SAML configuration
- [x] set up redirect from old domain
- [x] change site name in the Sites framework (through admin site)
- [ ] rename branch in deployment repository
- [x] the index prefix on elasticsearch should be updated to `textcavator-` for new corpora. Request updated API access
- [x] update matomo

### Related repositories
- [ ] ianalyzer-readers:
	- [ ] change repository name
	- [ ] update documentation
	- [ ] [change python package name](https://peps.python.org/pep-0423/#how-to-rename-a-project)
- [ ] various other repositories mention I-analyzer and may need to be updated. You can do an [organisation-wide query](https://github.com/search?q=org%3ACentreForDigitalHumanities+i-analyzer+-repo%3ACentreForDigitalHumanities%2FI-analyzer+-repo%3ACentreForDigitalHumanities%2Fianalyzer-readers+&type=code) to find them.
- [x] See what Zenodo does and if that needs attention

### Administration
- [x] rename folder in shared drive
- [x] rename teams channel + biweekly meeting

### PR
- [x] announcement in CDH newsletter
- [x] update CDH portfolio
- [x] communicate with stakeholders

Please add anything that's missing here! :)</div>

PR #1926, opened 2025-11-12
Replace "i-analyzer" -> textcavator

lukavdplas:
See #1925

## Changes to server configuration

Minor changes to environment/settings files to update the site name / title; these should also be updated for production environments.

After changing `SITE_NAME` in the backend, the site name must still be updated manually in the admin site.


backend/ianalyzer/settings.py
```diff
@@ -48,7 +48,7 @@
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 CSRF_TRUSTED_ORIGINS = ['http://localhost:8000']
 
-SITE_NAME = 'IANALYZER'
+SITE_NAME = 'Textcavator'
```
JeltevanBoheemen: This change requires a data migration, or manual change in the admin interface. See de7ea388014d4354de4788680f1f375ae1979bfc

PR #1930, opened 2025-11-19
Fix corpus refresh

### LINKED ISSUE #1917: Corpus form: refresh available corpora after activating corpus
When a user creates a corpus through the frontend form and activates it, it becomes available for them to search. But the frontend does not refresh the list of corpora available to the user so if go straight to the corpus, the interface claims that you are not authorised.

Solution: after activating/deactivating a corpus, re-request the list of corpora in the `CorpusService`.

lukavdplas:
Refreshes corpus data on relevant changes in the corpus form (close #1917)


frontend/src/app/corpus-definitions/form/index-form/index-form.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
 import { CorpusDefinition } from '@models/corpus-definition';
 import { APIIndexHealth, isComplete, JobStatus } from '@models/indexing';
-import { ApiService } from '@services';
+import { ApiService, CorpusService } from '@services';
```
JeltevanBoheemen: Seems like an unused import?

Issue #1954, opened 2025-12-05
Whitespace formatting in document content


lukavdplas:
The version currently in production (5.24) normalises the whitespace in document contents. This has a few effects:

- If you have a line like `A       B      C`, it will be shown as `A B C`.
- The text is split by linebreaks (one or more `\n`), and each line is wrapped in a paragraph `<p>` element. Paragraphs are semantic elements, but they are also styled with some extra spacing. (c.f. #1477)

I was looking at this while working on #1934. There is actually some code to preserve white space but it is not correctly implemented, so the whitespace is collapsed anyway.

Anyway, the advantage of doing it like this is that for many corpora, it makes the content more readable. And in many cases, splitting on linebreaks is semantically sound. I pushed this because it's better accessibility-wise, to actually make `<p>` elements when applicable.

"when applicable" being the key phrase, but that is not a valid assumption. The whitespace formatting looks good for corpora like Troonredes, but does not work for corpora like DBNL which also includes poems and songs.

Since Textcavator displays research data, I think it makes more sense to just display the whitespace as-is, and not add semantic encoding in the form of paragraph elements.

I should add: it's also possible to provide your text as an array of strings, which Textcavator interprets as an array of paragraphs. This behaviour seems fine to me, and like a decent way to encode paragraph breaks in the source data.

So a sensible system could be:
- Textcavator displays all whitespace as-is and does not "interpret" `\n` (by making it a semantic tag). 
- If you provide your text as an array, instead of a single string, those denote paragraphs, which are shown as `<p>` tags. Content within a paragraph is still displayed as-is.

But there are some hang-ups:

- Many existing corpora use a single `\n` between paragraphs. If you display the linebreaks as-is instead of adding paragraph margins, it's less readable.
- You currently cannot encode array values in the corpus form. I expect we'll add this at some point, tough, because it's also very useful for keyword fields. I don't think it's essential; if you want visually spaced out paragraphs, you can just use `\n\n`.

Possible solutions:
 
- Switch to the setup described above, but that means many existing corpora will look worse.
- Keep normalising whitespace, but then we need to accept that Textcavator is not suitable for some types of text (such as poetry), which is not great.
- Add a parameter for fields, which controls whether or not the whitespace should be normalised in the frontend, to support legacy data. This is probably the best outcome, but adds complexity.
[...]

PR #1959, opened 2025-12-09
Content formatting

### LINKED ISSUE #1934: DOM sanitisation in document contents
Currently, the logic to display document content fields includes a `bypassSecurityTrustHtml` call. This was necessary to insert icons for named entities, I believe, but it's a security vulnerability when we allow user-created corpora.

That said, the bypass does not seem to work at the moment: content is still sanitised. (And so the icons are not displayed.) 

Related issue: we don't support safe HTML tags for documents, but because the content is included with `[innerHTML]`, the tags are not escaped. ([Actually supporting HTML fields](https://github.com/CentreForDigitalHumanities/I-analyzer/discussions/1438) might be nice but would also require that we ignore tags in Elasticsearch.)

Proposed changes:
- Change the display logic to remove the bypass call.
- Tags within a document should be escaped.
- Fix the named entity view so it shows icons again.

lukavdplas:
This solves some minor issues and edge cases with regard to content formatting, and adds a lot of tests. Resolves #1934

- Document preview tables now use a fixed column width, so all previews are aligned.
- Content previews show linebreaks.
- All XML/HTML tags in document contents are now escaped.
- Fixed highlights not showing up when the text contains XML/HTML tags.
- Fixed content highlighting when the content is an array.
- Whitespace in documents is now shown as-is, and not collapsed. Paragraphs are still split for main content fields.

Unrelated: fixed a typo in one of the corpus categories.

Re. whitespace, see #1954 . This update keeps the paragraph normalisation but shows whitespace within paragraphs. In a future update, I'll probably add a boolean setting in the corpus field to control whether whitespace should be preserved or normalised.

### DIFF
--- a/frontend/src/app/document/content-field-preview/content-field-preview.component.html
+++ b/frontend/src/app/document/content-field-preview/content-field-preview.component.html
@@ -1,14 +1,12 @@
 @if (field && document) {
     <div class="break-word" style="white-space: pre-wrap;">
         @let snippets = document.highlight?.[field.name];
-        @if (snippets) {
-            @for (snippet of snippets; track $index; let last = $last) {
-                @for (par of snippet | paragraph:true; track par) {
-                    <p [class.mb-2]="!last" ia-highlighted-content [content]="par"></p>
-                }
-                @if (!last) { <hr> }
+        @for (snippet of snippets; track $index; let last = $last) {
+            @for (par of snippet | paragraph:true; track par) {
+                <p [class.mb-2]="!last" ia-highlighted-content [content]="par"></p>
             }
-        } @else {
+            @if (!last) { <hr> }
+        } @empty {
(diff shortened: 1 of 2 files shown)


frontend/src/app/document/pipes/keyword.pipe.ts
```diff
@@ -8,13 +8,9 @@ import * as _ from 'lodash';
...
-    transform(content: any): string {
-        if (_.isArray(content)) {
-            return content.map(String).join(', ');
-        } else {
-            return String(content);
-        }
-
+    transform(content: any): string[] {
+        const items = _.isArray(content) ? content : [content];
+        return items.map(String);
```
JeltevanBoheemen: This presumes `content` is mappable to String.
What happens if it's not?

PR #1962, opened 2025-12-16
index naming

### LINKED ISSUE #1159: Versioned indexing when alias is not set to latest version
Did something like this:

```bash
# make my-corpus-1 index and set alias
yarn django index my-corpus -p
yarn django alias my-corpus

# make my-corpus-2 index, but something goes wrong so I do not transfer the alias
yarn django index my-corpus -p

# try to make my-corpus-3 index
yarn django index my-corpus -p
```

The last operation will fail because the indexing function tries to make `my-corpus-2` a second time.

### LINKED ISSUE #1947: Indexing: no catch if intended alias is already used as index name
If you run `python manage.py index mycorpus --prod`, the setup function will look for any existing indices with the alias you would assign to your new index. It's designed to detect a numbered index named, for example, `textcavator-mycorpus-1` with alias `textcavator-mycorpus`.

The script will fail there is an index named `textcavator-mycorpus` (where this is the actual name and not the index). (Specifically, it will find the index and then assume that `textcavator-mycorpus` is the *alias* and not the name. Then it fails in a follow-up request because the alias does not exist.)

It's fine that the script fails in this scenario, but there should be a decent error message.

lukavdplas:
This makes some changes to how indices are named, and how `es_index` and `es_alias` are used.

The main reason was that the behaviour for database corpora needed to change, e.g. to allow duplicate names for database corpora in the future. I also wanted to change `es_alias` to work better for use cases like parlamint or course explorer.

- `es_index` is now optional, and intended to customise/override the default name. For Python corpora, the name is based on the corpus name, e.g. `textcavator-times`, for database corpora it's based on the ID, e.g. `textcavator-custom[3]`. Bracket characters are to avoid overlap with corpus names or version numbers. 
- `es_index` is no longer set when you create a corpus via the API (but you can still set it from the admin site). The generated index name is not derived from the corpus title, so this resolves https://github.com/CentreForDigitalHumanities/Textcavator/issues/1824
- Most test corpora no longer set `es_index`. Incidentally, that means you can use them in development too, without deleting the index every time you run unit tests. (Because the tests use a different application prefix.)
- `es_alias` now specifies an *additional* alias, rather than replacing the alias name. I think this is closer to how we use it. This means the corpus is still searchable under its own index name, as well as the alias.
- For a corpus with `es_alias`, the `index --prod --rollover` and `alias` commands now only remove the alias from earlier versions of the same corpus. (Not from indices that belong to other corpora.)
- Indexing with `--prod` fails with a clear message if you previously created an unversioned index for the corpus (close #1947)
- In production mode, you can now create a new index version when the active version is not the latest one. (close #1159)
- `es_index`/`es_alias` are no longer visible in the API.
- In some places, added a catch if the index name matches more than one index. 

As far as I know, this won't break anything for existing corpora, because they all still specify `es_index`.

### DIFF
--- a/backend/es/search.py
+++ b/backend/es/search.py
@@ -8,7 +8,7 @@ def get_index(corpus_name: str) -> str:
     if corpus.configuration.es_index:
         return corpus.configuration.es_index
 
-    config = settings.SERVERS.get(server_for_corpus(corpus.name))
+    config = settings.SERVERS[server_for_corpus(corpus.name)]
     prefix = config.get('index_prefix', None)
     name = corpus.name if corpus.has_python_definition else f'custom[{corpus.pk}]'
     return f'{prefix}-{name}' if prefix else name
--- a/backend/es/client.py
+++ b/backend/es/client.py
@@ -9,7 +9,7 @@ def elasticsearch(corpus_name):
     selected based on the CORPUS_SERVER_NAMES setting.
     '''
     server_name = server_for_corpus(corpus_name)
-    config = settings.SERVERS.get(server_name)
+    config = settings.SERVERS[server_name]
 
     return client_from_config(config)
 
--- a/backend/es/download.py
+++ b/backend/es/download.py
@@ -16,7 +16,7 @@ def scroll_chunks(corpus: str, query_model, download_size=None, client=None, **k
     index = get_index(corpus)
     if not client:
         client = elasticsearch(index)
-    server_conf = settings.SERVERS.get(server_for_corpus(corpus))
+    server_conf = settings.SERVERS[server_for_corpus(corpus)]
     scroll_timeout = server_conf['scroll_timeout']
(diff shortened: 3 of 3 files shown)


backend/es/search.py
```diff
@@ -1,17 +1,24 @@
 from typing import Dict
-from es.client import elasticsearch
-from addcorpus.models import CorpusConfiguration
+from es.client import elasticsearch, server_for_corpus, server_config
+from addcorpus.models import Corpus
 
-def get_index(corpus_name):
-    corpus_conf = CorpusConfiguration.objects.get(corpus__name=corpus_name)
-    return corpus_conf.es_index
+def get_index(corpus_name: str) -> str:
```
JeltevanBoheemen:
```suggestion
def get_index(corpus_name: str) -> Optional[str]:
```
Seems this can return `None`. Or should it raise in that case?

PR #1971, opened 2025-12-18
iframe test app

lukavdplas:
Creates a small application that can be used to test the embedded view for Textcavator locally. Thought this would be convenient for Peace Portal and similar projects in the future.

The app is a Wordpress-vibes webpage with the Troonredes corpus:

screenshot of website with textcavator search page embedded


JeltevanBoheemen (commented):
I like the code, and I like what it allows us to test.

However from an informal discussion we have had about this PR:
The extra top-level directory makes me uncomfortable. I think we discussed maybe releasing this as a separate repo? Do you have notes on our conversation @lukavdplas ? Going off the to of my head here.

lukavdplas:
I agree the top-level directory is odd, it just didn't make sense inside an existing directory. Separate repo sounds fine too!

lukavdplas:
I'll close this and make a new repo then :+1:

JeltevanBoheemen:
Be aware we might need `X_FRAME_OPTIONS = 'ALLOW'` in this repo.

lukavdplas:
Done -> https://github.com/CentreForDigitalHumanities/textcavator-iframe-dev

PR #1975, opened 2026-01-05
ParlaMint 5

Meesch:
At long last, it is here, all 29 ParlaMint corpora! This PR includes:
 - A big parlamint-all corpus definition that describes all the fields in `parlamint_all.py`
 - 29 separate child-corpora in `parlamint_subcorpora.py`
 - In the `parlamint_utils` there are three auxiliary files: one for constants, one for the extract logic, and one for the transform logic. This is all done to keep the sizes of these files somewhat under control.
 - Some additional stopword lists were added from sources outside of nltk for languages that nltk does not support

When reviewing, it is most useful to be able to test the index locally, so if you want the test data, I can send those over. The main files to review are `parlamint_all.py` and `parlamint_subcorpora.py`.

### DIFF
--- a/backend/corpora/parliament/finland.py
+++ b/backend/corpora/parliament/finland.py
@@ -8,7 +8,7 @@ from corpora.parliament.parliament import Parliament
 import corpora.parliament.utils.field_defaults as field_defaults
 from corpora.utils.constants import document_context
 from bs4 import BeautifulSoup
-from corpora.parliament.utils.parlamint_v4 import extract_all_party_data, extract_people_data, extract_role_data, party_attribute_extractor, person_attribute_extractor, clean_value
+from corpora.parliament.utils.parlamint import extract_all_party_data, extract_people_data, extract_role_data, party_attribute_extractor, person_attribute_extractor, clean_value
 
 from django.conf import settings
 
--- a/backend/corpora/parliament/utils/parlamint.py
+++ b/backend/corpora/parliament/utils/parlamint.py
@@ -1,0 +1,232 @@
+from ianalyzer_readers.extract import XML, Combined, Metadata
+from bs4.element import NavigableString
+from bs4.element import NavigableString, Tag as Node
+from string import punctuation
+from typing import Iterable
+
+from addcorpus.es_mappings import non_indexed_text_mapping, keyword_mapping
+from ianalyzer_readers.xml_tag import Tag
+from addcorpus.python_corpora.filters import MultipleChoiceFilter
+from addcorpus.python_corpora.corpus import FieldDefinition
+
+"""
+This file remains in place, but it is outdated. More recent parlamint logic can be found in 
+backend/corpora/parliament/clarin_parlamint/parlamint_utils. This file is kept for backwards 
+compatibility for the Finland corpus.
+"""
+
+
+def clean_value(value):
+    if type(value) == str or type(value) == NavigableString:
+        stripped = value.strip()
+        if len(stripped):
+            return stripped
+    if type(value) == int or type(value) == float:
+        return value
+
+
+def extract_party_data(node):
+    id = node['xml:id']
+
+    full_name_node = node.find('orgName', full='yes')
+    full_name = full_name_node.text if full_name_node else None
+
+    abbreviation_node = node.find('orgName', full='init')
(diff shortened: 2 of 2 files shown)


backend/addcorpus/es_settings.py
```diff
@@ -23,32 +23,42 @@ def get_language_key(language_code):
...
+    nltk_path = _nltk_stopwords_path(language_code)
+    supplementary_path = _supplementary_path(language_code)
+    if os.path.exists(nltk_path):
+        with open(nltk_path) as infile:
+            words = [line.strip() for line in infile.readlines()]
+            return words
+    elif os.path.exists(supplementary_path):
+        with open(supplementary_path) as infile:
             words = [line.strip() for line in infile.readlines()]
             return words
```
lukavdplas: To avoid repetition:

```python
    if os.path.exists(nltk_path):
        return _read_stopwords_file(nltk_path)
    elif os.path.exists(supplementary_path):
        return _read_stopwords_file(supplementary_path)
    # else: ...

def _read_stopwords_file(path: str) -> List[str]:
    with open(path) as infile:
        return [line.strip() for line in infile.readlines()]
```

backend/corpora/parliament/clarin_parlamint/images/parlamint_NL.jpg
lukavdplas: Kind of nitpicky but some of these images are oversized for thumbnails (e.g. this one is 6000 x 3500 pixels). Smaller images are nicer for users with mobile data / slow network / etc.

backend/corpora/parliament/clarin_parlamint/parlamint_all.py
```diff
@@ -0,0 +1,336 @@
...
+        soup = BeautifulSoup(f, features="xml")
+    return soup
+
+class ParlaMintAll(XMLCorpusDefinition):
+    title = "All ParlaMint Corpora (version 5.0)"
+    description = "All corpora from the ParlaMint dataset, including 27 countries"
+    category = "parliament"
+    image = "parlamint.png"
+    description_page = "parlamint_all.md"
+    languages = ['en']
```
lukavdplas: Isn't this mixed-language? Or is it only the translations?

backend/corpora/parliament/clarin_parlamint/parlamint_all.py
```diff
@@ -0,0 +1,336 @@
...
+    data_directory = settings.PARLAMINT_ALL_DATA
+    visualize = []
```
lukavdplas: I assume you copied this from another class, but it doesn't do anything. (I think it was a parameter years ago, so that's why it's still in some old corpus classes.)

```suggestion
```

backend/corpora/parliament/clarin_parlamint/parlamint_all.py
```diff
@@ -0,0 +1,336 @@
...
+    max_date = datetime(year=2022, month=12, day=31)
+    
+    document_context = document_context()
+    tag_toplevel = Tag('TEI')
+    tag_entry = Tag('u')
+
+    country_codes = COUNTRY_CODES
+    country_code = None
+
+    def sources(self, *args, **kwargs):
```
lukavdplas: I thought the plan was to use a shared alias for the combined corpus?

backend/corpora/parliament/clarin_parlamint/parlamint_all.py
```diff
@@ -0,0 +1,336 @@
...
+        display_type='text_content',
+        visualizations=['wordcloud', 'ngram'],
+        csv_core=True,
+        language=None,
+    )
+    speech.extractor = XML(
+            Tag('s'),
+            multiple=True,
+            extract_soup_func = extract_speech,
+            transform=' '.join)
```
lukavdplas: The same extractor is also defined in `parlamint_subcorpora.py`

lukavdplas (commented):
Good work!

As we already discussed, I think it would be better if we used a shared alias for the ParliamentAll corpus, and there could be less repetition in the subcorpus definitions.

Other than that, this looks good. I added some minor comments here. The only real issue is the changes to `corpora/parliament/utils/field_defaults.py`, which would affect the P&P corpora as well.

PR #1994, opened 2026-01-29
fix document position in search popup

### LINKED ISSUE #1992: Document popup shows incorrect index
### What went wrong?

The document popup in the search results shows the index of the document in the header, but it's not correct. The count resets at every page but the total suggests it doesn't.

E.g., the first document on page 2 may show "document 1 of 223600", which should be "document 21 of 223600".

See above. Correct count would be `page_index * 20 + document_index + 1`

### Screenshot

Image

- [x] https://textcavator.hum.uu.nl
- [ ] a local server

### Version

5.26.0

### Steps to reproduce

- Open search page for any corpus
- Go to page 2 of the results
- Use "view" for a document.

lukavdplas:
close #1992

### DIFF
--- a/frontend/src/app/document/document-popup/document-popup.component.html
+++ b/frontend/src/app/document/document-popup/document-popup.component.html
@@ -4,7 +4,7 @@
         [dismissableMask]="true"
         [blockScroll]="true">
         <ng-template  #header>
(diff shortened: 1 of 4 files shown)


JeltevanBoheemen (approved):
Huh, never noticed this bug. Nice one.

PR #2007, opened 2026-02-12
Clarify required fields

lukavdplas:
Clearer workflow in the corpus form:
- For several fields, the form now validates they are filled in before saving.
- Form shows validation messages.
- Form labels use "*" for required fields.
- Added aria-labelledby / aria-describedby for some form inputs
- Disabled form steps have grey text (which is clearer)
- When the next step in the form is disabled, shows an explanatory message at the bottom instead of a disabled button
- Stricter typing for some `FormGroup`s.

### DIFF
--- a/frontend/src/app/corpus-definitions/form/field-form/field-form.component.ts
+++ b/frontend/src/app/corpus-definitions/form/field-form/field-form.component.ts
@@ -212,9 +212,8 @@ export class FieldFormComponent implements OnChanges {
         this.dialogService.showManualPage('types-of-fields');
     }
 
-    languageLabel(field: FormGroup): string {
-        const value = field.controls.language.value;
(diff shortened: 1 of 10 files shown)


frontend/src/app/corpus-definitions/form/field-form/field-form.component.ts
```diff
@@ -222,6 +236,11 @@ export class FieldFormComponent implements OnChanges {
         this.fieldsForm.updateValueAndValidity();
     }
 
+    private onValidationFail() {
+        this.fieldsForm.markAllAsTouched();
+        this.validationFailed$.next();
```
JeltevanBoheemen: Why do all fields need to be marked as toucher here?

PR #2010, opened 2026-02-16
Batch termvectors requests

lukavdplas:
This improves the computation time somewhat for the ngrams visualisation, and the term frequency visualisation if you use wildcard search. Uses a batched request for document vectors, so the backend only makes 1 request per 100 documents instead of requesting each document individually. 

(This doesn't seem to make any difference on my local server, but it did make a difference when I tried it on our test server.)


lukavdplas:
Re. speed tests: I time these tasks when I work on features like this, but mostly by hand. We could automate it by writing a script, but then I'm not sure what we would really do with it.

Re. computation times in general, I agree this is worth improving, though there is often a trade-off with code complexity or storage (or both).

PR #2015, opened 2026-02-24
Language analysis

lukavdplas:
Based on #2012 and #1975 , I want to make some changes to the way we configure language analysis so we can add more detailed customisation per language.

The old system was to infer the configuration by checking available stopword lists and stemmer filters, but this is quite limited. It's a problem for languages very dissimilar to English, but languages like German and French would also benefit from using ES's built-in filters `german_normalization` and `french_elision` respectively.

Another problem was that we had no definitive list of "supported" languages. Amongst other problems, this made it hard to match input to the list of known languages, e.g. for equivalent codes like `nl` and `nld`.

So with the new system, we need to explicitly add a configuration for each language. The upside is that those configurations can override every aspect of text analysis. I've also added some documentation on how text analysis is set up.

I have also cherry-picked the extra stopword lists from #1975.


backend/addcorpus/language_analysis.py
```diff
@@ -0,0 +1,697 @@
...
+    by other modules, e.g. to get the stopwords list.
+
+    Miminum required attributes:
+        code: the IETF tag for the language.
+        has_stopwords: indicates support for stopwords filtering
+        has_stemming: indiciates support for stemming in addition to stopword filtering
+
+    If you set `has_stopwords` or `has_stemming` to `True`, the class will generate
+    analyzers based on the NLTK stopwords corpus, a simple number filter, and built-in
+    stemmers from Elasticsearch. You may need to override these methods, e.g. to use a
```
Meesch: This is correct, but it might be even clearer to also mention that there are no NLTK stopword lists or ES stemmers for several languages, in which case you will have to use a custom list or stemmer.

backend/addcorpus/language_analysis.py
```diff
@@ -0,0 +1,697 @@
...
+
+        This removes stopwords and stems tokens. Will be used in a 'stemmed' multifield.
+        '''
+        if self.has_stemming:
+            analyzer = self._clean_analyzer()
+            analyzer['filter'].append(self._stemmer_filter_name)
+            return analyzer
+
+
+# LANGUAGE SPECS
```
Meesch: Probably a matter of taste, but I would keep a separate file for the subclasses per language, and call it something like analyzers_per_language.py. Makes this file more general and a bit more static, whereas the subclasses file can be updated more regularly.

backend/addcorpus/language_analysis.py
```diff
@@ -0,0 +1,697 @@
...
+
+    def _clean_analyzer(self):
+        analyzer = super()._clean_analyzer()
+        analyzer['filter'] = [
+            'catalan_elision', 'lowercase', self._stopwords_filter_name,
+        ]
+        return analyzer
+
+
+class Chinese(LanguageAnalyzer):
```
Meesch: Hmm, this is an analyzer for 'general Chinese', which is a [macrolanguage](https://en.wikipedia.org/wiki/ISO_639_macrolanguage#zho), meaning that it can have several sub-codes for certain alphabet types/character systems. Could we ask the Chinese researcher which type of Chinese is present in the corpus? The question I have is: should we create subclasses for this Chinese class, or should we have them as completely separate classes if we have a corpus with `cmn` (Mandarin) or `yue` (Cantonese)?

backend/addcorpus/language_analyzers.py
```diff
@@ -0,0 +1,697 @@
...
+        analyzer = super()._clean_analyzer()
+        analyzer['filter'] = [
+            'italian_elision', 'lowercase', self._stopwords_filter_name
+        ]
+        return analyzer
+
+    _stemmer_filter_language = 'light_italian'
+
+
+class Latvian(LanguageAnalyzer):
```
Meesch, line 308: To a lesser extent Latvian is also a macrolanguage designation, but it seems unlikely that we will ever have a corpus in [Latgalian](https://en.wikipedia.org/wiki/Latgalian_language) (though I still included this for your exploration pleasure)

backend/addcorpus/language_analyzers.py
```diff
@@ -0,0 +1,697 @@
...
+    code = 'lv'
+    has_stopwords = True
+    has_stemming = True
+
+    _stopwords_source = 'supplementary'
+
+
+class NorwegianBokmal(LanguageAnalyzer):
```
Meesch, line 351: Aha, my previous question is answered here: different classes for different language forms, no subclasses. I think this method is fine, I do not think we necesessarily need to organize the analyzers hierarchically as it does make it even more fluffy and layered.  If we want to make this our standard, let's adopt it in the docs.

backend/addcorpus/language_analysis.py
lukavdplas: Yes it's probably clearer to change the code to cmn - that is how I handled the codes for Norwegian (also a macrolanguage). It doesnt matter much, functionally. If you would add an extra analyzer for cantonese, then either way, a field tagged as zh or cmn will get this analyzer, and a field tagged yue would get the cantonese analyzer.

PR #2024, opened 2026-03-03
Wordcloud tokenisation

### LINKED ISSUE #2023: Use ES tokenizer in wordcloud
For the purpose of searching and most visualisations, documents are tokenised by Elasticsearch, but the wordcloud uses its own tokeniser (using [CountVectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html) ). This is more primitive than the one used by Elasticsearch, and does not work well for all languages.

Use the termvectors API to get token counts from Elasticsearch.

- You can also pass a custom tokeniser function to the CountVectorizer, but then we would need to reverse-engineer the tokeniser we already have.
- You can turn off the wordcloud for some languages (admittedly not a great loss to science). This is what I did as a quick fix in #2012

lukavdplas:
close #2023 

This requests the tokenised text for the wordcloud from Elasticsearch, which is more consistent than using the `CountVectorizer`.


Meesch (approved):
Nice work!

PR #2047, opened 2026-03-26
handle invalid directory in save_corpus

lukavdplas:
Changes the way `loadcorpora` handles invalid data directory values: this will now raise a warning and save a blank value, instead of breaking the corpus.


JeltevanBoheemen:
What problem does this solve?

lukavdplas:
The present validation is too strict. If a corpus does not have a valid data directory, it's a problem during indexing, but it should not stop you from using an already index corpus.

This can be inconvenient locally, but the main issue is on the production server, when it cannot access the data directory used by the index server.

Issue #1796, opened 2025-05-14
Corpus form: JSON definition workflow


lukavdplas:
The proposed workout for #1795 would mean that the corpus form requires that you upload CSV data as the basis for the corpus, before you can continue with further steps.

This is kind of at odds with the option to base your corpus on a JSON definition. Technically, you could still do that: create the corpus from the JSON file, upload an image in the metadata form (but leave the rest untouched), upload compatible CSV data, skip the field configuration, index the corpus. But this is not very intuitive or clear.

One problem is that the JSON does not contain an image or CSV data, so in the step-based form, later steps are blocked. At the same time, if you are using a JSON definition as a basis, most of the form can be ignored, yet it's not clear the user what is required at this point.

Another problem is that uploading CSV data will normally generate a list of fields which can be further configured in the next step, but if the user is starting with a definition file, they have already specified exactly what fields they want. Yet they won't be able to verify that these fields were imported correctly before uploading a CSV file, which may overwrite the fields.  

### Solution 1: dual workflow

This would mean there is a different "mode" for JSON-based corpus creation (which will probably reuse many of the same components). Here the steps would be:
 
- Upload JSON file and image
- Upload CSV data and verify compatibility with the corpus definition
- Index/publish

By contrast, the form-based corpus creation would be:

- Fill in metadata and upload image
- Upload CSV data, which generates fields
- Tweak generated fields
- Index/publish

That said, the primary purpose for the JSON option is to allow back-ups, transferring corpora between servers, etc.: I don't imagine any end-users would prefer editing a corpus this way.

### Solution 2: upload everything up front

Another solution is that the "import" option requires that the user upload the JSON, image, and CSV in a single form, so they are submitted (and validated) together. Once everything is uploaded, the corpus is created and the user is taken to the step-based form. All steps are now unlocked as if the user has gone through the full form, and the user can go straight to indexing.

PR #2048, opened 2026-03-26
Update corpus import docs

lukavdplas:
- Make the first-time setup docs more straightforward by instructing the user to import one of the test corpora.
- The "basic" test corpus is now called `ExampleCorpus` instead of `MockCSVCorpus`.
- Removed some documentation about adding corpora from JSON files. Adding this as developer documentation only made sense when the UI  wasn't developed yet.


backend/addcorpus/tests/test_word_models_present.py
JeltevanBoheemen, line 1: In this file, I would keep the `Corpus` part of example classes.
I know they are consumed directly below, but if in the future we would acces them elsewhere it's better to be explicit what the example constitutes.

documentation/Adding-existing-corpora.md
```diff
@@ -0,0 +1,19 @@
...
+
+The source files of a corpus are not included in this directory; ask another developer about their availability. If you have (a sample of) the source files for a corpus, you can add the corpus your our environment as follows:
+
+1. Add the corpus to the `CORPORA` dictionary in your local settings file. See [CORPORA settings documentation](./Django-project-settings.md#corpora).
+2. Set custom settings for your corpus. Check the definition file to see which variables it expects to find in the Django settings. Some of these may be optional.
+3. Activate your python virtual environment. Run the `loadcorpora` admin command (`yarn django loadcorpora`) to register the new corpus in the SQL database. Then create an ElasticSearch index from the source files by running, e.g., `yarn django index mycorpus`. See [Indexing](./Indexing-corpora.md) for more information.
+
+### Database-only corpora
+
+Database-only corpora are still in development, but we plan to support making copies of your created corpora by downloading the configuration. This feature should be documented in the user manual.
```
JeltevanBoheemen: Exclude this until it's implemented

JeltevanBoheemen (changes requested):

documentation/Adding-existing-corpora.md
lukavdplas: It is already implemented! But the workflow is not great (#1796)

This documentation is not very clear though, I'll update it.

PR #2077, opened 2026-05-27
Feature/hansard 2026

Meesch:
Nieuwe UK Data


backend/corpora/parliament/uk-new.py
```diff
@@ -0,0 +1,376 @@
...
+from addcorpus.python_corpora.corpus import XMLCorpusDefinition, FieldDefinition
+from addcorpus.python_corpora.filters import MultipleChoiceFilter
+from addcorpus.es_mappings import keyword_mapping, text_mapping, date_mapping, main_content_mapping
+from corpora.parliament.parliament import Parliament
+import corpora.parliament.utils.field_defaults as field_defaults
+from corpora.utils.constants import document_context
+
+
+
+def extract_date(path: str):
```
JeltevanBoheemen, line 24:
```suggestion
def extract_date(path: str) -> Optional[str]:
```
Type annotations are not consistent in this file. Not a big deal, but if you're doing them anyway, might as well make them complete.

Annotating all parameters types has higher prio then return types imho.

backend/corpora/parliament/uk-new.py
```diff
@@ -0,0 +1,376 @@
...
+    else:
+        return None
+
+def extract_chamber(path: str):
+    if 'daylord' in path:
+        return 'House of Lords'
+    elif 'debates' in path:
+        return 'House of Commons'
+    else:
+        return None
```
JeltevanBoheemen, line 37: Very nitpicky, but noticed this in more functions in this file. `else` clause and `return None` is reduntant, though I would argue explicitly returning `None` is a ncie thing to do.
```suggestion
    if 'daylord' in path:
        return 'House of Lords'
    elif 'debates' in path:
        return 'House of Commons'
    return None
```

backend/corpora/parliament/uk-new.py
```diff
@@ -0,0 +1,376 @@
...
+                speaker_ids.append(tag.attrs['person_id'].split('/')[-1])
+
+    return speaker_ids
+
+def select_topic(input):
+        full_speech_id, topic_dict = input
+        speech_id = abbreviate_speech_id(full_speech_id)
+        previous_topic = ''
+        for key in topic_dict:
+            if float(key) > float(speech_id):
```
JeltevanBoheemen, line 87: Smart way to compare `10.6` to `10.7` :)

backend/corpora/parliament/uk-new.py
```diff
@@ -0,0 +1,376 @@
...
+                return previous_topic
+            else:
+                previous_topic = topic_dict[key]
+
+def lookup_person_attribute(lookup_tuple):
+    metadata_dict, id, name, label = lookup_tuple #name is only included for debugging purposes
+
+    id = id.split('/')[-1] if id else None # twfy ID is at the end of uri
```
JeltevanBoheemen, line 95: What is `twfy`?

frontend/src/app/filter/multiple-choice-filter/multiple-choice-filter.component.html
```diff
@@ -1,8 +1,16 @@
...
+        [id]="filter.corpusField.name" 
+        [disabled]="!options.length" 
+        [filter]="options.length>=6" 
+        [options]="options" 
+        [virtualScroll]="true"
+        [virtualScrollItemSize]="60"
+        [maxSelectedLabels]=1
+        placeholder="Choose" 
+        [ngModel]="data" 
+        (onChange)="update($event.value)" ariaLabelledBy="legend-filter-{{filter.displayName | slugify}}" fluid>
```
JeltevanBoheemen, line 13: This seems to work well (tested on acc), but is a separate feature. 
The filters are well tested. Should the virtualscroll be tested there? (not sure, may be fine like this).

JeltevanBoheemen (commented):
Well done, seems like quite smoe work to figure out the data cleaning.

Remark on the corpus/file name: using `new` to indicate a version is not future proof. What happens when we get another update? We will end up with things like `Index_Current_New` and `Index_Current_Old` (actual example from another project). Maybe call it `uk_hansard`?

Some nitpicky code quality comments, feel free to ignore if you consider them too small to fix :)

backend/corpora/parliament/uk-new.py
Meesch: they work for you

PR #2113, opened 2026-07-14
Feature/angular 21

lukavdplas:
Updates Angular to 21

Node requirements are the same, so this shouldn't cause any issues with deployment.

Some notable changes:
- This also includes an automatic update for template control flow syntax, e.g. `*ngIf` to `@if`. Hence the massive number of changes.
- This update also mangled the whitespace for HTML files. So be it. (But perhaps a linting PR might be in order later.)
- Some components had to be changed to standalone to deal with build errors.


JeltevanBoheemen (approved):
Test, build and serve work fine.
Don't make me review the code please :)

PR #2114, opened 2026-07-14
Feature/unify settings formats

JeltevanBoheemen:
Reverts the `get_deprecated_setting` helper, and adjusts all corpora using it to the new settings format.

Also removes (almost) all old-style settings (`getatttr(settings.SOME_SETTING, <default>`). Exceptions:
- Course Explorer related corpora
- Peace Portal related corpora 

This was a combination of manual moving of settings, and using an LLM-based helper. So a quick sanity check would be much appreciated.

### Deployment
Branches updated for:
- [textcavator](https://github.com/CentreForDigitalHumanities/deployment-rsl/commit/94b1f053be574139a7d3b248b1b512ac970e7b81)
- [people&parliament](https://github.com/CentreForDigitalHumanities/deployment-rsl/commit/71ae0b7b0ef0acae4ec25375ce5eca59d2ad9057) 

Sanity checks also appreciated

### DIFF
--- a/backend/corpora/guardianobserver/guardianobserver.py
+++ b/backend/corpora/guardianobserver/guardianobserver.py
@@ -3,6 +3,7 @@ Collect information from the Guardian-Observer corpus: the articles are containe
 separate xml-files, zipped.
 '''
 
+from addcorpus.python_corpora.corpus import XMLCorpusDefinition, FieldDefinition
 import logging
 logger = logging.getLogger(__name__)
 import re
@@ -18,7 +19,6 @@ from ianalyzer_readers import extract
 from api.utils import find_media_file
 from indexing.run_update_task import update_document
 from addcorpus.python_corpora import filters
-from addcorpus.python_corpora.corpus import XMLCorpusDefinition, FieldDefinition, get_deprecated_setting
 from media.image_processing import sizeof_fmt
 from media.media_url import media_url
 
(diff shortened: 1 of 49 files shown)


backend/corpora/guardianobserver/guardianobserver.py
```diff
@@ -36,28 +36,14 @@ class GuardianObserver(XMLCorpusDefinition):
...
     max_date = datetime(year=2003, month=12, day=31)
-
-    @property
-    def data_directory(self):
-        return get_deprecated_setting('GO_DATA') or super().data_directory
-
-    @property
-    def es_index(self):
-        return get_deprecated_setting('GO_ES_INDEX') or 'guardianobserver'
+    scan_image_type = 'application/pdf'
```
lukavdplas, line 39:
```suggestion
    scan_image_type = 'application/pdf'
    es_index = 'guardianobserver'
```

backend/corpora/parliament/netherlands.py
```diff
@@ -428,27 +428,12 @@ class ParliamentNetherlands(Parliament, XMLCorpusDefinition):
...
     category = "parliament"
     document_context = document_context()
 
-    @property
-    def data_directory(self):
-        return get_deprecated_setting('PP_NL_RECENT_DATA') or super().data_directory
-
-    @property
-    def data_directory_old(self):
-        return get_deprecated_setting('PP_NL_DATA')
```
lukavdplas, line 442: `data_directory_old` is now missing from the class. You can just set it to `None`, but it looks like there is no implementation for an empty value, so I think this should be:

```py
@property
def data_directory_old(self):
    raise NotImplementedError('Required configuration data_directory_old is missing')
```

backend/ianalyzer/settings_test.py
```diff
@@ -16,10 +16,15 @@ def mock_corpus_path(*path):
...
-UBLAD_DATA = '' # necessary to make ublad test not fail
+CORPUS_SETTINGS = {
+    "times": {
+        "data_directory": os.path.join(BASE_DIR, 'addcorpus', 'python_corpora', 'tests'),
+        "es_index": "test-times",
+    },
+    "ublad": {
+        "data_directory": ""
+    }
+}
```
lukavdplas, line 23: The ublad setting isn't needed anymore. This would have been required to import the module in a test:

https://github.com/CentreForDigitalHumanities/Textcavator/blob/477e0393ff6ecc246a097caa20c55d20687f8b0d/backend/corpora/ublad/tests/test_ublad.py#L3

The `ublad` module would raise an AttributeError on import if `settings.UBLAD_DATA` does not exist. But now you can just import the module, yay!

lukavdplas (changes requested):
Nice! A few minor changes required.

backend/corpora/parliament/netherlands.py
JeltevanBoheemen: I sort of assumed the old implementation was deprecated, but I'll include this to be backward compatible.

lukavdplas: It's just that `data_directory_old` is still required in the settings (if you want to index), but it's good if the class points to that somehow.

PR #2126, opened 2026-07-29
Bootstrap

lukavdplas:
Mega-update to switch from bulma to bootstrap. From here we can also start replacing primeNG components, but I have not started with that yet.

(PrimeNG components still work in the meantime, so they can be replaced one by one.)

This branch is more or less ready, but there are still going to be some small stray issues before it's release-ready. But I think the code should be stable enough that we can pick up different issues in parallel.

Summary of changes:
- Replaced class names and restructured templates throughout the application.
- `DropdownComponent` now builds on `NgbDropdown`. NgbDropdown handles opening/closing the menu, our own component only needs to manage the current selection. API is slightly changed.
- `MenuDropdownComponent` (for navbar dropdowns) is replaced by `NgbDropdown`.
- `TabsComponent` is replaced by `NgbNav`.
- `BalloonDirective` is replaced by `NgbTooltip`.
- `ToggleComponent` is replaced by the `switch` CSS class.
-  The `is-loading` CSS class is replaced by `LoadingDirective`.
- For buttons, the `is-loading` CSS class is replaced by `ButtonLoadingDirective`. This one also blocks <kbd>Enter</kbd>/<kbd>Space</kbd> events while loading. (The CSS class only blocked pointer events.)
- `ConfirmModalComponent` now uses `NgbModal`; `PaginationComponent` uses `NgbPagination`; `NotificationsComponent` uses  `NgbToast`.
- The global CSS now includes a `tc_theme.scss` that imports bootstrap with customisations. This is meant to separate the theme from the generic `styles.scss`, and should make it easier to provide different theme colours in the future.

Still to do:

- There are probably some components I forgot to update, layout issues to address, etc.

PrimeNG components to replace:

- `p-dialog` -> replace with `NgbModal`
- `p-datepicker` -> replace with `NgbDatepicker`. 
- `p-steps` -> use `NgbNav` (with extra styling) -> #2139 
- `p-autocomplete` for query comparison -> build from scratch -> #2141 
- `p-autoComplete` for language selection -> similar, but mix in `NgbTypeahead` somehow
- `p-multiselect` -> Could build this directly on `NgbDropdown` (without a generic "multiselect" component), or build something analogous to our (single-select) `DoprdownComponent`.
- `p-slider` -> there may be a suitable package for multi-thumb bootstrap sliders. Otherwise build from scratch.
- `p-table` -> implement filters and pagination (with `NgbPagination`) without a "table component".


frontend/src/app/search/search.component.ts
```diff
@@ -89,6 +89,7 @@ export class SearchComponent implements OnInit, OnDestroy {
     }
 
     onTabChange(tab: SearchTab) {
+        console.log(tab);
```
Meesch, line 92: 👀

frontend/src/app/shared/dropdown/dropdown.component.ts
```diff
@@ -18,7 +18,43 @@ import { DropdownService } from './dropdown.service';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { NgbDropdown } from '@ng-bootstrap/ng-bootstrap';
 
-
+/**
```
Meesch, line 21: I appreciate the documentation for this component being moved to the component, I think this is a good place for it, but I do wonder if it might be useful to have a README.md file for the dropdown functionality, since it does have quite some moving parts that could maybe most easily be described in a seperate file that describes the links between component, module, service, and directives?

frontend/src/app/core/menu/menu.component.ts
```diff
@@ -37,6 +37,8 @@ export class MenuComponent implements OnDestroy, OnInit {
         { label: 'dark', icon: themeIcons.dark, value: Theme.DARK },
     ];
 
+    /** active theme (light/dark) */
+    activeTheme$ = this.themeService.theme$;
```
Meesch, line 41: maybe add an explanation that this is necessary because the menu is on permanent dark mode?

frontend/src/tc_theme.scss
```diff
@@ -0,0 +1,113 @@
...
+$link-color-dark: tint-color($primary, 50%);
+$primary-border-subtle-dark: tint-color($primary, 10%);
+
+// custom fonts
+$font-family-sans-serif: fonts.$tc_font_sans_serif;
+$display-font-family: fonts.$tc_font_brand;
+$display-font-weight: normal;
+
+// bold weight for form labels
+$form-label-font-weight: bold;
```
Meesch, line 42: I think I prefer the form labels to be closer to their original styling: not bold and a little smaller. I will add it to the list of things for the general review.

frontend/src/app/download/download.component.html
```diff
@@ -1,112 +1,102 @@
...
         Only the first {{downloadLimit}} results will be included in the file.
       </p>
     </div>
   }
   <form>
-    <div class="field">
-      <label class="label" id="fields-select-label">Fields</label>
-      <div class="control">
+    <div class="mb-3">
+      <label class="form-label" id="fields-select-label">Fields</label>
```
Meesch, line 16: You removed the stale control div in commit 67de89d12fbf6c719826b5199becbfd3a7065152, but the removal of the div means that the multiselect appears right next to the 'Fields' label/header, instead of on a newline.

Meesch (commented):
Great work! Very well-modularized commits, and all functionalities remain functional on my machine. 

I really like these improvements:
 - overall structuring and code cleanliness
 - pagination widget looks better
 - highlighting looks much better

I have noticed the following points:
 - The new bootstrap theme uses bold large lettering for the filter headers (`form-label`) and corpus headers (`text-bg-primary`), I think I prefer the previous font styling, but we can discuss.
 - when selecting a filter (any filter), the selection colour is green, not blue
 - we have lost some of the width of the app: it would be a shame to give up our newfound wideness. My hunch is that it has something to do with the `column` class in Bootstrap, I think it has extra padding in it.
 - The dark mode is less dark :/


What is still needed for this to be merged? Maybe decide on styling decisions next meeting?

lukavdplas:
Thanks! Re. the status: there is some polishing required but I would say: let's hold off with merging so we can squeeze in a release next week or so. After that, we can merge it and fix remaining issues on develop.

> The new bootstrap theme uses bold large lettering for the filter headers (form-label) and corpus headers (text-bg-primary), I think I prefer the previous font styling, but we can discuss.

Agreed :)

> When selecting a filter (any filter), the selection colour is green, not blue

These are primeNG component that are now using their default primary colour. We used to customise this, but bootstrap did not have convenient CSS variables. We can make the variables or replace the components before release.

> we have lost some of the width of the app: it would be a shame to give up our newfound wideness.

Agreed, we can customise that.

> The dark mode is less dark :/

Didn't notice tbh. We can customise the colours, though (as we did with Bulma).

XanderVertegaal:
I'll pick up `p-dialog` and `p-datepicker` before moving to others.

PR #2142, opened 2026-08-19
Add T-value option for collocations

### LINKED ISSUE #2105: ngrams: new "compensate for frequency" formula
In the ngram graph, the formula used then "compensate for frequency" is selected should be changed. New formula will be based on PPMI which is more conventional.

lukavdplas:
Adds a new option for calculating collocation scores based on the T-value. Also significantly expands the documentation for the neighbouring words graph.

T-values seem to work in providing a more "balanced" score, compared to the old formula. Results seem reasonable to me but I'll let the people & parliament team test it after release.

Some comments:

- An issue with these formulas is that we have to take a lot of shortcuts with the statistics (now also documented in the manual), e.g. taking background frequencies from the entire corpus instead of the query set. So the validity is... questionable.
- The  t-value formula is also available in n-grams mode, but whether that is really valid, I don't know.


JeltevanBoheemen (approved):
Code is clean and sensible, explanation in the HTML is clear.

